PR #49, opened 2020-12-15
Fix heading levels

chaance:
The main thing here is that `Subtitle` should not be an `h2`, and all heading levels should be adjusted accordingly for proper semantics. I suspect we should change the default `Subtitle` in the DS because this is a good general rule to follow. Here's a blog that references why using headings for subtitles is usually not the best way to go: https://gomakethings.com/how-to-create-accessible-subtitles/

The as prop typing in the DS is also broken, which I've worked around by casting the value as `any` for now. Related issue: https://github.com/modulz/design-system/issues/15

This is a draft at the moment because I've only updated `Accordion` headings for now. Styles should be completely identical as a result of this change.


components/MDXComponents.tsx
```diff
@@ -44,17 +44,16 @@ const LinkHeading = ({
 
 export const MDXComponents = {
   h1: (props) => <DS.Title {...props} css={{ mb: '$1', ...props.css }} />,
-  h2: (props) => <DS.Subtitle {...props} css={{ mt: '$2', mb: '$6', ...props.css }} />,
-  h3: ({ children, id, ...props }) => (
+  h2: ({ children, id, ...props }) => (
     <LinkHeading id={id} css={{ mt: '$7', mb: '$2', ...props.css }}>
-      <DS.Heading {...props} id={id} data-heading style={{ scrollMarginTop: '45px' }}>
+      <DS.Heading {...props} as={'h2' as any} id={id} data-heading style={{ scrollMarginTop: '45px' }}>
```
benoitgrelard, line 49: I know that's not new but I'm confused about the `css` prop coming in. Reading the code here, it looks to me that if we had a `css` prop passed in (somehow) it would be added to both the `LinkHeading` AND the `DS.Heading` which is probably not what we expect.

@peduarte would the css prop ever even be passed through these considering they are the mdx mapped components? How would one do that?

peduarte: yeah, true. its a poor refactor. I think we should only forward css to `LinkHeading`

peduarte (approved):

components/MDXComponents.tsx
benoitgrelard: Is it possible to pass a `css` prop though? Considering these work by doing:

```md
## I'm a heading…
```

PR #63, opened 2020-12-16
HTML validation tweaks

chaance:
- `dd` needs to be a direct child of `dl`
- There were instances of block elements nested in inline elements, so I cleaned those up as well
- Added accessible text to the `ExternalIcon` component since they convey meaning

### DIFF
--- a/components/ExternalIcon.tsx
+++ b/components/ExternalIcon.tsx
@@ -1,12 +1,26 @@
-export function ExternalIcon() {
+import { VisuallyHidden } from '@radix-ui/react-visually-hidden';
+import * as React from 'react';
+
+export function ExternalIcon(props: React.ComponentPropsWithoutRef<'svg'>) {
   return (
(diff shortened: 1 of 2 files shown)


components/ExternalIcon.tsx
```diff
@@ -1,12 +1,26 @@
-export function ExternalIcon() {
+import { VisuallyHidden } from '@radix-ui/react-visually-hidden';
+import * as React from 'react';
+
+export function ExternalIcon(props: React.ComponentPropsWithoutRef<'svg'>) {
```
benoitgrelard, line 4: Can we use our `AccessibleIcon` primitive for this?

chaance: Yeah, I think we can do that in several places actually. Can I do this as a follow-up PR just to keep them simple and knock out several others in one sweep?

benoitgrelard: Works for me!

benoitgrelard (approved):

PR #66, opened 2020-12-16
Use accessible-icon primitive

chaance:
@benoitgrelard This one's for you, homie

### DIFF
--- a/components/DocsPage.tsx
+++ b/components/DocsPage.tsx
@@ -8,7 +8,7 @@ import { RadixLogo } from './RadixLogo';
 import { PrimitivesNav } from './PrimitivesNav';
 import { DesignSystemNav } from './DesignSystemNav';
 import { useProductType } from '../utils/useProductType';
-import { VisuallyHidden } from '@radix-ui/react-visually-hidden';
+import { AccessibleIcon } from '@radix-ui/react-accessible-icon';
 
 export function DocsPage({ children }: { children: React.ReactNode }) {
   const router = useRouter();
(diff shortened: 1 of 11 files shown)


components/DocsPage.tsx
```diff
@@ -65,8 +65,7 @@ export function DocsPage({ children }: { children: React.ReactNode }) {
                   ':focus': { boxShadow: 'none' },
                 }}
               >
-                <VisuallyHidden>Radix Documentation (Alpha)</VisuallyHidden>
-                <RadixLogo aria-hidden />
+                <RadixLogo label="Radix Homepage" />
```
benoitgrelard, line 68: The label seemed different here, are we ok with changing it?

pages/blog/index.tsx
```diff
@@ -33,22 +33,7 @@ export default function Home() {
...
-                overflow: 'hidden',
-                clip: 'rect(0, 0, 0, 0)',
-                whiteSpace: 'nowrap',
-                border: 0,
-              }}
-            >
-              Stitches homepage
-            </span>
-            <RadixLogo />
+            <RadixLogo label="Radix Homepage" />
```
benoitgrelard, line 36: Label has changed, probably wasn't right, but may not be radix homepage?

pages/design-system/index.tsx
```diff
@@ -19,22 +19,7 @@ export default function DesignSystemHome() {
...
-                overflow: 'hidden',
-                clip: 'rect(0, 0, 0, 0)',
-                whiteSpace: 'nowrap',
-                border: 0,
-              }}
-            >
-              Stitches homepage
-            </span>
-            <RadixLogo />
+            <RadixLogo label="Radix Homepage" />
```
benoitgrelard, line 22: Same question for label

components/DocsPage.tsx
chaance: Yeah, it was wrong before. It should be the label for the link, which goes to the homepage.

pages/blog/index.tsx
chaance: I think it is. Right now everything is redirected to the Radix intro page which is defacto home. We can change these later as needed.

PR #68, opened 2020-12-16
Fix scrollbar jank on QuickNav (#64)

### LINKED ISSUE #64: Scroll locking elements make the layout shift
This is most likely due to the right aside being position fixed

dropdown

Recorded on Firefox Windows, confirmed same behavior on chrome Windows as well, testing on the linux ones later today

chaance:
Fixes #64 by adding the helper class from [react-remove-scroll](https://github.com/theKashey/react-remove-scroll#positionfixed-elements) designed to deal with `position: fixed` elements when `right` is set to `0`.

### DIFF
--- a/layouts/DocsLayout.tsx
+++ b/layouts/DocsLayout.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import NextLink from 'next/link';
 import { useRouter } from 'next/router';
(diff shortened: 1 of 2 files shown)


benoitgrelard (approved):
Nice, I knew it was to do with position fixed. Not great that the abstraction leaks, we should be able to do better if we create our own component/api to prevent scroll.

PR #75, opened 2020-12-17
Improve accessibility section of menu components

benoitgrelard:
This pull request:

- [ ] Fixes a bug
- [ ] Other

### DIFF
--- a/pages/primitives/docs/components/context-menu.mdx
+++ b/pages/primitives/docs/components/context-menu.mdx
@@ -1055,6 +1055,8 @@ export default () => (
 
 ## Accessibility
 
+Uses [roving tabindex](https://www.w3.org/TR/wai-aria-practices-1.2/#kbd_roving_tabindex) to manage focus movement among menu items.
+
 ### Keyboard Interactions
 
 <KeyboardTable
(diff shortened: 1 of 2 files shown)


peduarte (approved):

pages/primitives/docs/components/context-menu.mdx
```diff
@@ -1078,11 +1080,7 @@ export default () => (
     },
     {
       keys: ['Esc'],
-      description: (
-        <span>
-          Closes <Code>Content</Code>.
-        </span>
-      ),
+      description: 'Closes the context menu',
```
chaance, line 1083: @benoitgrelard I know this was merged already and it's not a super big deal, but I think it's probably wiser to use JSX formatting in these spots by default even if we don't actually need them. Took me ages to refactor all of the ones that needed it before when we decided to use nested `Code` tags, for example. Seems pointless to need to switch back and forth when we change content IMO.

benoitgrelard: haha I kept inline with the current convention which seemed to be string when only string is needed, jsx otherwise. I don't have very strong opinions though on this though 😀

PR #82, opened 2021-01-21
Version switcher

peduarte:
Version switcher 🆒 

Key changes:
- each primitive can now contain multiple versions of docs
- the URL structure is: `/primitives/docs/components/accordion/0.0.1`
- the root URL points to the latest version: `/primitives/docs/components/accordion` ⇢ `/primitives/docs/components/accordion/0.0.1`
- there is no manual maintenance of rewriting the root URL to the latest version, this happens automatically
- a version switcher UI, driven by a native `select` element has been added
- link to npm has been moved
- the folder structure for primitive components and utilities must be: `/primitiveName/primitiveVersion/index.mdx`
- the `version` field in the front matter has been removed as this can now be derived from the folder structure

image


layouts/DocsLayout.tsx
```diff
@@ -351,98 +355,95 @@ function QuickNav() {
...
+            <Box as="span" css={{ ml: '$1', color: '$gray700', position: 'absolute', right: -20 }}>
+              <ExternalIcon />
+            </Box>
+          </Flex>
+        </Link>
+      </Box>
+      <Box css={{ mb: '$2' }}>
+        <Link
+          variant="blue"
           href={`https://www.npmjs.com/package/@radix-ui/react-${name}`}
```
benoitgrelard, line 398: Didn't we say we'd point this to the correct version on npm?

next.config.js
```diff
@@ -55,6 +72,45 @@ module.exports = withPlugins(
...
+    // So navigating to /tooltip rewrites to /tooltip/[latestVersion]
+    async rewrites() {
+      const isDirectory = (fileName) => fs.lstatSync(fileName).isDirectory();
+      const getPaths = (dir) =>
+        fs
+          .readdirSync(dir)
+          .map((file) => path.join(dir, file))
+          .filter(isDirectory);
+
+      const getPathOfLatestVersion = (dir) =>
```
benoitgrelard, line 86: I'm confused by `dir` here. The input is not a directory right? it should be an array of paths if I understand correctly.

next.config.js
```diff
@@ -55,6 +72,45 @@ module.exports = withPlugins(
...
+
+      const primitivesDirectory = path.join(__dirname, 'pages/primitives/docs/components');
+      const primitivesPaths = getPaths(primitivesDirectory);
+      const latestPrimitivesPaths = getPathOfLatestVersion(primitivesPaths);
+
+      const utilitiesDirectory = path.join(__dirname, 'pages/primitives/docs/utilities');
+      const utilitiesPaths = getPaths(utilitiesDirectory);
+      const latestUtilitiesPaths = getPathOfLatestVersion(utilitiesPaths);
+
+      const all = [...latestPrimitivesPaths, ...latestUtilitiesPaths];
```
benoitgrelard, line 105: I think you can just skip that `all` variable and inline it, because I don't think `all` is very descriptive anyway…

next.config.js
```diff
@@ -55,6 +72,45 @@ module.exports = withPlugins(
...
+
+      return all.reduce((acc, curr, index) => {
```
benoitgrelard, line 107: Would be good to name `acc, curr` a bit better, like `redirectMap, path`, something like that.

next.config.js
```diff
@@ -55,6 +72,45 @@ module.exports = withPlugins(
...
+        const [, destination] = all[index].split('/pages');
+        const [, source] = path.join(all[index], '..').split('/pages');
```
benoitgrelard, line 109: you don't need `index` right? `all[index]` is the same as `curr`, it's your path.

benoitgrelard (commented):
This is great! 👏 ♥️ 

I've left a few minor comments

layouts/DocsLayout.tsx
peduarte: we can't point to the right version on github, since theyre not being tagged on release. so to keep it consistent I left it out too.

next.config.js
peduarte: yeah, agree. was useful for loggin only

layouts/DocsLayout.tsx
benoitgrelard: fair enough, once we can do both we can edit this.

next.config.js
peduarte: yeah that works

next.config.js
peduarte: yeah lets use `path` 👍

next.config.js
peduarte: true, thats better

colmtuite:
Awesome 👏 

It needs a more pronounced hover state. I'd drop the hover background-color in favour of a stroke, so `boxShadow: inset 0 0 0 1px $gray600`. Then a 5px left margin to compensate to add some space between the control and label.

colmtuite (approved):

benoitgrelard (approved):

peduarte:
@colmtuite good idea, done!

image

benoitgrelard:
🚀

PR #88, opened 2021-01-27
New release: update docs

benoitgrelard:
Published a new version of the packages with the latest fixes, updated docs accordingly here.


peduarte (approved):
LGTM but its quite hard to review

I wonder if we can find a better way to diff the versions. Lets chat on a call

benoitgrelard:
Sorry I should have mentioned but when I do a single PR it means that the brand new files are pure duplicates of previous, so nothing to review on these.

Otherwise I'd do 2 PR's like I did the other day, where first one just duplicates, and second shows changes.

PR #104, opened 2021-03-02
[Tooltip][Slot] Rename props prefixed with `is`

jjenzz:
I noticed some of the code examples on `Tooltip` and `Slot` were using the old APIs.


benoitgrelard (commented):
Probably need to update in all older versions too as I think that change was made before initial release.

PR #107, opened 2021-03-04
Update data attributes

jjenzz:
Remove `data-radix-*` references from docs where no longer relevant.


benoitgrelard (commented):
Looks good!

I would just swap the order of these 2 sections now:

image

benoitgrelard (approved):

PR #111, opened 2021-03-05
Update docs – create duplicated versions

benoitgrelard:
Just upgraded to new release, duplicated files + removed `live` from previous version.


package.json
```diff
@@ -10,32 +10,31 @@
   },
   "dependencies": {
     "@modulz/design-system": "0.0.15",
-    "@modulz/radix-icons": "^3.3.0",
```
benoitgrelard, line 13: This wasn't used as it has long be replaced by `@radix-ui/react-icons`

jjenzz (approved):

PR #117, opened 2021-03-15
Add disabled trigger example in tooltip docs

benoitgrelard:

This pull request:

- [ ] Fixes a bug
- [ ] Other

### DIFF
--- a/pages/primitives/docs/components/tooltip/0.0.9/index.mdx
+++ b/pages/primitives/docs/components/tooltip/0.0.9/index.mdx
@@ -345,7 +345,9 @@ const StyledContent = styled(Tooltip.Content, {
 
 ### With disabled trigger
 
-Because a disabled trigger will not fire events, adding a tooltip on a disabled trigger require a little bit more work. You will need to render the trigger as a wrapper (`span` for example) and add your disabled button inside. Furthermore, the disabled button will need a few more styles.
+A disabled button will not fire events so adding a tooltip to a disabled trigger requires a little more work.
+
+Render the trigger as a wrapper (`span` for example) and place your disabled button inside with some CSS to disable pointer events. This allows the tooltip to receive all mouse and keyboard events instead of being blocked by the button.
 
 ```jsx live manual
 // import { styled } from 'path-to/stitches.config';
--- a/pages/primitives/docs/components/tooltip/0.0.9/index.mdx
+++ b/pages/primitives/docs/components/tooltip/0.0.9/index.mdx
@@ -341,6 +341,48 @@ const StyledContent = styled(Tooltip.Content, {
 });
 ```
 
+## Examples
+
+### With disabled trigger
+
+Because a disabled trigger will not fire events, adding a tooltip on a disabled trigger require a little bit more work. You will need to render the trigger as a wrapper (`span` for example) and add your disabled button inside. Furthermore, the disabled button will need a few more styles.
+
+```jsx live manual
+// import { styled } from 'path-to/stitches.config';
+import * as Tooltip from '@radix-ui/react-tooltip';
+import { BlendingModeIcon } from '@radix-ui/react-icons';
+
+const StyledContent = styled(Tooltip.Content, {
+  borderRadius: 3,
+  padding: '5px 10px',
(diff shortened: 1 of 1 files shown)


pages/primitives/docs/components/tooltip/0.0.9/index.mdx
```diff
@@ -341,6 +341,48 @@ const StyledContent = styled(Tooltip.Content, {
 });
 ```
 
+## Examples
+
+### With disabled trigger
+
+Because a disabled trigger will not fire events, adding a tooltip on a disabled trigger require a little bit more work. You will need to render the trigger as a wrapper (`span` for example) and add your disabled button inside. Furthermore, the disabled button will need a few more styles.
```
jjenzz, line 348:
```suggestion
Because a disabled trigger will not fire events, adding a tooltip on a disabled trigger requires a little bit more work. You will need to render the trigger as a wrapper (`span` for example) with your disabled button inside and add a few styles.
```

pages/primitives/docs/components/tooltip/0.0.9/index.mdx
```diff
@@ -341,6 +341,48 @@ const StyledContent = styled(Tooltip.Content, {
...
+  fill: 'gainsboro',
+});
+
+export default () => (
+  <Tooltip.Root>
+    <Tooltip.Trigger as="span">
+      <button
+        type="button"
+        disabled
+        style={{ pointerEvents: 'none' }}
```
jjenzz, line 373: Are these the "few" styles? I'm wondering why they need to add `pointerEvents: 'none'` to a disabled button... aren't they already disabled on disabled buttons?

jjenzz (changes requested):

pages/primitives/docs/components/tooltip/0.0.9/index.mdx
benoitgrelard: Not by default no, it's likely some design systems would add this rule to disabled buttons, but we can't be sure.

jjenzz: Oooh I realised why I've gotten mixed up. I was confused cos I was thinking `disabled` already prevents `onClick` etc, which is why we needed this workaround in the first place. But I realise now this style is to prevent CSS pseudo states?

Aren't CSS pseudo states completely unrelated to the issue we're describing a fix for though? Like, Tooltips aside, if I did `<button disabled>` I would still get hover states etc. it has nothing to do with getting tooltips to work, just a consumer style they can decide how to handle. For example, I used to do this:

```jsx
const Button = styled('button', {
  '&:enabled:hover': {}
  // etc...
});
```

benoitgrelard: No it does, it is necessary to make tooltips work on disabled buttons.

jjenzz: I'm so confused then. Can we explain in the docs how that change makes tooltips work?

benoitgrelard: It's because the trigger which is now a wrapper has exactly the same size as the button inside it, and so `mouseleave` won't ever trigger. Adding `pointerEvents: 'none'` fixes that because now it's has if the disabled button wasn't there.

jjenzz: Ooooh i seeeeeeeee.... okay, yeah that makes perfect sense. Totally forgot we need tooltip to receive the other events 😂  

In that case, can we maybe update the text to read something like:

> A disabled button will not fire events so adding a tooltip to a disabled trigger requires a little more work. 
>
> Render the trigger as a wrapper (`span` for example) and place your disabled button inside with some CSS to disable pointer events. This allows the tooltip to receive all mouse and keyboard events instead of being blocked by the button.

?

benoitgrelard: Sounds great, I'll update.

jjenzz (approved):

PR #120, opened 2021-03-25
Update docs - part 1

benoitgrelard:
This PR does a few things:
- Upgrade primitives versions to latest
- Create a file for each package's latest version (note the gap here cause of all the bad versions in between)
- Remove the `live` from codeblocks on previous version
- Rename all the things for the `toggle-button => toggle` name change

Still to come next:
- Document new custom duration props for `Tooltip`
- Document new `ToggleGroup` primitive
- Document new `Toolbar` primitive

PR #121, opened 2021-03-25
Update docs - part 2 (Custom timings on Tooltip)

benoitgrelard:
----------

- [x] Change base branch to `main` once #120 is merged. 

----------

### DIFF
--- a/pages/primitives/docs/components/tooltip/0.0.15/index.mdx
+++ b/pages/primitives/docs/components/tooltip/0.0.15/index.mdx
@@ -402,12 +402,19 @@ export default () => (
 
 ### Show instantly
 
-Use the `delayDuration` to control the time it takes for the tooltip to open. You can define these per-instance, but it is best to set a default value for the prop so all your tooltips use the same timings.
+Use the `delayDuration` to control the time it takes for the tooltip to open. We recommend wrapping the tooltip in your own component with a default value for the prop for consistent timings across your app. You will still be able to override the prop per-instance.
+
+You can use our `extendPrimitive` utility for this.
 
 ```jsx live manual compact
 // import { styled } from 'path-to/stitches.config';
 import * as Tooltip from '@radix-ui/react-tooltip';
+import { extendPrimitive } from '@radix-ui/react-primitive';
 import { ZoomInIcon, ZoomOutIcon } from '@radix-ui/react-icons';
+
+const MyTooltip = extendPrimitive(Tooltip.Root, {
+  defaultProps: { delayDuration: 0 },
+});
 
 const StyledContent = styled(Tooltip.Content, {
   borderRadius: 3,
@@ -423,7 +430,7 @@ const StyledArrow = styled(Tooltip.Arrow, {
 
 export default () => (
   <>
-    <Tooltip.Root delayDuration={0}>
+    <MyTooltip>
       <Tooltip.Trigger>
         <ZoomInIcon />
       </Tooltip.Trigger>
@@ -431,9 +438,9 @@ export default () => (
         Zoom In
         <StyledArrow />
       </StyledContent>
-    </Tooltip.Root>
+    </MyTooltip>
 
-    <Tooltip.Root delayDuration={0}>
+    <MyTooltip>
       <Tooltip.Trigger style={{ marginLeft: 10 }}>
         <ZoomOutIcon />
       </Tooltip.Trigger>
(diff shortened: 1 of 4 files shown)


pages/primitives/docs/components/tooltip/0.0.15/index.mdx
```diff
@@ -118,6 +118,20 @@ Contains all the parts of a tooltip.
         </span>
       ),
     },
+    {
+      name: 'delayDuration',
+      type: 'number',
+      default: 700,
+      description:
+        'The duration from when the mouse enters the trigger until the tooltip gets opened.',
```
peduarte, line 126: > tooltip gets opened.

that's a bit odd 🤔 

how about:
- until the tooltip opens
- until the tooltip has opened
- until the tooltip is opened

pages/primitives/docs/components/tooltip/0.0.15/index.mdx
```diff
@@ -385,6 +399,54 @@ export default () => (
...
+      <Tooltip.Trigger>
+        <ZoomInIcon />
+      </Tooltip.Trigger>
+      <StyledContent>
+        Zoom In
+        <StyledArrow />
+      </StyledContent>
+    </Tooltip.Root>
+
+    <Tooltip.Root delayDuration={0}>
```
peduarte, line 437: should this be `skipDelayDuration`?

pages/primitives/docs/components/tooltip/0.0.15/index.mdx
benoitgrelard: "until the tooltip opens" nice

pages/primitives/docs/components/tooltip/0.0.15/index.mdx
benoitgrelard: No why?

peduarte: Oh, ok. I thought the correlation between the content and the code block was weird. 

> Use the `delayDuration` and `skipDelayDuration` to control the timings at which a tooltip is shown.

Youre talking about two different props here

> These will always show instantly:

Saying that "these" will show instantly, I thought you each button corresponds to each of the props above

then the code block shows two identical tooltips but only demoing one of the props

benoitgrelard: Yeah I didn't want to confuse too much about showing what the other prop did, but just mentioning it as it's related to controlling "timings". Any ideas on how to make this less confusing?

peduarte: I think one example for each, or be explicit that the example there is only about `delayDuration`

benoitgrelard: I've ended up reducing this to just an example of the `delayDuration` prop which is by far the most common thing people want to control.

peduarte (approved):

pages/primitives/docs/components/tooltip/0.0.15/index.mdx
```diff
@@ -385,6 +400,52 @@ export default () => (
 );
 ```
 
+### Show instantly
+
+Use the `delayDuration` to control the time it takes for the tooltip to open. You can define these per-instance, but it is best to set a default value for the prop so all your tooltips use the same timings.
```
jjenzz, line 405: > but it is best to set a default value for the prop so all your tooltips use the same timings

I read this part and then was confused when the example doesn't show how to do this. Do you think we should show an example with `const Tooltip = extendPrimitive(Tooltip.Root, { delayDuration: 400 });`?

pages/primitives/docs/components/tooltip/0.0.15/index.mdx
```diff
@@ -403,6 +464,10 @@ export default () => (
       keys: ['Enter'],
       description: 'If open, closes the tooltip without delay.',
     },
+    {
+      keys: ['Escape'],
+      description: 'If open, closes the tooltip without delay.',
+    },
```
jjenzz, line 470: Nice catch :+1:

pages/primitives/docs/components/tooltip/0.0.15/index.mdx
benoitgrelard: I did think about this but wasn't sure we wanted to push that thing yet as we're not 100% sure about it right?

PR #132, opened 2021-04-12
Initial implementation

peduarte:

This pull request:

- [ ] Fixes a bug
- [ ] Other


colmtuite:
I made some changes.

**Fixes**

- Extended the mini nav to reach the top and bottom of the viewport.
- Tweaked the MDX components to improve the type setting.
- Made the Link2Icon darker.
- Began redesigning the Releases page. This is incomplete. Ping me when you’re looking at this and we’ll work together.

**Questions / Issues**

- Why do we need this new mdx-remote thing? What is it giving us?
- When a new page loads, the content is flashing and janking. It looks like content is being lazy-loaded. It’s especially noticeable if you scroll down the current page before you navigate to a new page.
- Navigating to a new page takes ages on localhost.
- When you open a Dialog, DropdownMenu, ContextMenu etc., the scrollbar causes the page to shift. I see something about a helper class but it doesn’t appear to be working?
- Where are the old DS docs gone? We had some rough ones before?
- Why are the QuickNav container styles duplicated in 3 places? Can’t we reuse one component?
- The new “Show Code“ button setup is noisy/anxiety-inducing, takes up more space, and doesn’t provide a way to collapse the code panel. We should switch back to the previous approach.

jjenzz:
I was just clearing up my Github Notifications and spotted that the release page is broken https://website-git-mdx-remote-radix-ui.vercel.app/primitives/docs/overview/releases

Just a heads up before this is merged 🙂

benoitgrelard:
@jjenzz it seems @colmtuite is working on that page as far as I can tell. Colm, maybe we should do this in a separate PR? This PR is already insanely huge and as to what I can see, it looks like you may want to introduce quite big changes to that release page which would necessitate componentizing as it's not making use of built-in markdown syntax.

peduarte:
Colm let's go through it today on our call

colmtuite (approved):

PR #137, opened 2021-04-20
Add preload to font

peduarte:
Thanks @StephenHaney 🥳

### DIFF
--- a/pages/_document.tsx
+++ b/pages/_document.tsx
@@ -11,7 +11,12 @@ export default class Document extends NextDocument {
           <style id="stitches" dangerouslySetInnerHTML={{ __html: getCssString() }} />
           <script dangerouslySetInnerHTML={{ __html: renderSnippet() }} />
           <link rel="icon" href="/favicon.png" />
-          <link rel="stylesheet" href="https://develop.modulz.app/fonts/fonts.css" />
+          <link
+            href="https://develop.modulz.app/fonts/fonts.css"
+            rel="preload"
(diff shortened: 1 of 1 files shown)


pages/_document.tsx
```diff
@@ -11,7 +11,12 @@ export default class Document extends NextDocument {
           <style id="stitches" dangerouslySetInnerHTML={{ __html: getCssString() }} />
           <script dangerouslySetInnerHTML={{ __html: renderSnippet() }} />
           <link rel="icon" href="/favicon.png" />
-          <link rel="stylesheet" href="https://develop.modulz.app/fonts/fonts.css" />
+          <link
+            href="https://develop.modulz.app/fonts/fonts.css"
+            rel="preload"
+            as="font"
+            crossOrigin="anonymous"
+          />
```
jjenzz, line 19: I know this is already merged, but there's a few things here confusing me:

- You've said `as="font"` but you're actually pulling in a stylesheet. 
- I thought for `preload` we're supposed add the preload `link` and the thing we're preloading, as it's just a hint for things to come. 
	- https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content
	- https://ashton.codes/preload-google-fonts-using-resource-hints/

Is this PR actually working as intended? I need an explainer here if so 😅

peduarte: Thanks Jenna, you're right. This has been fixed here https://github.com/modulz/stitches-site/pull/73

PR #142, opened 2021-04-23
Migrate to `mdx-bundler`

peduarte:
This PR replaces `next-mdx-remote` with `mdx-bundler`

Reasons for switching:
- `next-mdx-remote` uses delays the rendering of the MDX content
- `next-mdx-remote` is really slow in dev mode. pages take >5s to build sometimes
- `next-mdx-remote` inlines lots of data into the HTML

A related thread https://twitter.com/peduarte/status/1385248989549838337

image


components/MDXComponents.tsx
```diff
@@ -33,14 +33,17 @@ import * as toggleDemos from './demos/Toggle';
...
+import { Frontmatter } from 'types/frontmatter';
 
 export const components = {
   ...DS,
   h1: (props) => (
     <DS.Text {...props} as="h1" size="8" css={{ fontWeight: 500, mb: '$2', lineHeight: '40px' }} />
   ),
-  Description: (props) => <DS.Subtitle {...props} as="p" css={{ mt: '$2', mb: '$7' }} />,
+  Description: ({ children, ...props }) => {
+    const childText = typeof children === 'string' ? children : children.props.children;
```
benoitgrelard, line 44: I find this very confusing to read, we've pulled `children` from `props` and now we do `children.props.children`

lib/rehype-meta-attribute.js
```diff
@@ -0,0 +1,23 @@
...
+    var match;
+
+    if (node.tagName === 'code' && node.data && node.data.meta) {
+      re.lastIndex = 0; // Reset regex.
+
+      while ((match = re.exec(node.data.meta))) {
+        node.properties[match[1]] = match[2] || match[3] || match[4] || '';
+      }
+    }
+  }
```
benoitgrelard, line 22: What does this new plugin do?

peduarte: There's a link at the top, but it converts meta to attributes. so doing ` ```jsx collapsed` passes `collapsed` as a prop to the mdx component

components/MDXComponents.tsx
peduarte: Yeah, that's different now because of how the [xdm compiler](https://github.com/wooorm/xdm) works, which is slightly different from the mdx one. [More info](https://github.com/wooorm/xdm/issues/47)

But basically that ensures that there's never a `p` inside `Description`. 

I'll add a comment there

benoitgrelard: Interesting… that's pretty weird! haha

peduarte: yeah its basically:

```
This is a paragraph

<div>This is a text node in a div</div>

<div>
  This is a paragraph in a div (aka, back in markdown)
</div>
```

PR #148, opened 2021-05-03
03/05/2021 release – part 3

benoitgrelard:
> 🔥 This PR builds on top of #147, don't merge directly

- update docs for all existing components
- new HoverCard docs will be in next PR

PR #149, opened 2021-05-03
03/05/2021 release – part 4

benoitgrelard:
> 🔥 This PR builds on top of #148, don't merge directly

- add new HoverCard docs


components/releaseHelpers.tsx
```diff
@@ -18,7 +18,7 @@ export function PRLink({ id }: { id: number | number[] }) {
     <Text css={{ display: 'inline', fontSize: '$2' }}>
       –{' '}
       {ids.map((id, i, arr) => (
-        <>
+        <React.Fragment key={id}>
```
peduarte, line 21: Why adding key in a fragment instead of on the `Link`?

peduarte (approved):

components/releaseHelpers.tsx
benoitgrelard: Because that's the node that needs keying, putting it on the `Link` wouldn't work.

peduarte: Ah cool, because there's a sibling. Didnt see that

PR #162, opened 2021-05-20
[RadioGroup] Remove name from PropsTable

benoitgrelard:
The name prop is no longer necessary as the name is propagated from the RadioGroup.Root component as explained by @benoitgrelard  in this [closed issue](https://github.com/radix-ui/primitives/issues/655)


benoitgrelard:
Thanks for this @Guardiadelcastro, I just took your change from your patch and PR'd it in as I think you just forgot to do that, but brilliant!

PR #164, opened 2021-05-24
Remove casting from polymorphic types

jjenzz:
Updates polymorphic docs to remove type casting recommendation. It seems some of the changes I've been making recently have meant we no longer need to cast. 

This approach will hopefully feel more natural to consumers as people tend to prefer to avoid casting.

Thanks @benoitgrelard for pointing out that we can do this without issue.

### DIFF
--- a/data/primitives/utilities/polymorphic/0.0.11.mdx
+++ b/data/primitives/utilities/polymorphic/0.0.11.mdx
@@ -42,9 +42,11 @@ Make a polymorphic `Box` component.
 import React from 'react';
 import type * as Polymorphic from '@radix-ui/react-polymorphic';
 
-const Box = React.forwardRef(({ as: Comp = 'div', ...props }, forwardedRef) => (
+type PolymorphicBox = Polymorphic.ForwardRefComponent<'div', {}>;
+
+const Box: PolymorphicBox = React.forwardRef(({ as: Comp = 'div', ...props }, forwardedRef) => (
   <Comp {...props} ref={forwardedRef} />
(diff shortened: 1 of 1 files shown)


benoitgrelard (approved):

data/primitives/utilities/polymorphic/0.0.11.mdx
```diff
@@ -56,7 +58,7 @@ export default () => <Box>
 
 ### ForwardRefComponent
 
-Cast your `forwardRef` component to this type to enable polymorphism.
+Types a `forwardRef` component as a polymorphic component.
```
benoitgrelard, line 61:
```suggestion
Types a `forwardRef` component as a polymorphic.
```

Perhaps?

jjenzz: Changed to read:

```
Adds polymorphic `as` prop types to a `forwardRef` component.
```

Is that clearer do you think?

data/primitives/docs/utilities/polymorphic/0.0.10.mdx
```diff
@@ -0,0 +1,148 @@
+---
```
benoitgrelard, line 1: I'm confused why there's a brand new file here?

peduarte (approved):
Interesting. This has a big impact on the DS as it's currently all done with casting. Does it have a performance impact?

benoitgrelard:
> Interesting. This has a big impact on the DS as it's currently all done with casting. Is this approach better for performance?

No big impact on DS as it's not a required change, mostly just a readability change.

PR #167, opened 2021-06-03
Fix broken "Edit this page on Github" url

andrevenancio:
This PR fixes the bug where the wrong URL would be generated for the Edit this page on Github button. Since the `docs` is inherited by the folder structure inside the `pages/docs` we need to replace it from the url.

There was also a initial value for editUrl which contains a `undefined.mdx` on the path since `currentPageSlug` is undefined. I think its more readable having the variable set to empty string, but this is arguable :)

### DIFF
--- a/components/PrimitivesPage.tsx
+++ b/components/PrimitivesPage.tsx
@@ -19,7 +19,7 @@ export function PrimitivesPage({ children }: { children: React.ReactNode }) {
 
   const GITHUB_URL = 'https://github.com';
   const REPO_NAME = 'radix-ui/website';
-  let editUrl;
+  let editUrl: string = '';
 
   if (typeof routerSlug === 'string') {
     currentPageSlug = router.pathname.substr(1).replace('[slug]', routerSlug);
--- a/components/PrimitivesPage.tsx
+++ b/components/PrimitivesPage.tsx
@@ -19,7 +19,7 @@ export function PrimitivesPage({ children }: { children: React.ReactNode }) {
 
   const GITHUB_URL = 'https://github.com';
   const REPO_NAME = 'radix-ui/website';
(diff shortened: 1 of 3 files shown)


components/PrimitivesPage.tsx
```diff
@@ -227,7 +227,7 @@ export function PrimitivesPage({ children }: { children: React.ReactNode }) {
         <Container size="3" css={{ maxWidth: '780px', my: '$9' }}>
           <Text size="3">
             <Link
-              href={editUrl}
+              href={editUrl.replace('docs/', '')}
```
peduarte, line 230: I'm not a fan of having the `.replace` here. I'd prefer this to mutation to happen higher up, where we're building this string.

For example:
```diff
if (typeof routerSlug === 'string') {
  currentPageSlug = router.pathname.substr(1).replace('[slug]', routerSlug);
  editUrl = `${GITHUB_URL}/${REPO_NAME}/edit/main/data/${currentPageSlug}.mdx`;
} else {
  currentPageSlug = router.pathname.substr(1).replace('[...slug]', routerSlug[0]);
  editUrl = `${GITHUB_URL}/${REPO_NAME}/edit/main/data/${currentPageSlug}/${routerSlug[1]}.mdx`;
}

+ editUrl = editUrl.replace('/docs', '');
```

peduarte (changes requested):
Hey! Thanks for this 🙂 

Left a small comment. 

> There was also a initial value for editUrl which contains a undefined.mdx on the path since currentPageSlug is undefined. I think its more readable having the variable set to empty string, but this is arguable :)

Agreed!

peduarte (approved):

peduarte:
Nice one! Thanks :)

PR #173, opened 2021-06-15
Add release notes

benoitgrelard:


andy-hook (approved):
Great write up man, looks good from my end.

PR #176, opened 2021-06-15
[Tabs] Update docs

benoitgrelard:


data/primitives/components/tabs/0.0.14.mdx
```diff
@@ -189,7 +189,7 @@ Contains all the tabs component parts.
 
 ### List
 
-Contains the tabs aligned along the edge of the active panel.
+Contains the triggers aligned along the edge of the active content.
```
andy-hook, line 192:
```suggestion
Contains the triggers that are aligned along the edge of the active content.
```

Not sure here, but the original sentence doesn't read super well for me.

data/primitives/components/tabs/0.0.14.mdx
```diff
@@ -1,15 +1,15 @@
 ---
 metaTitle: Tabs
-metaDescription: A set of layered sections of content—known as tab panels—that display one panel of content at a time.
+metaDescription: A set of layered sections of content—known as tab panels—that are displayed one at a time.
```
andy-hook, line 3: Just curious why these are hyphenated?

andy-hook (approved):

data/primitives/components/tabs/0.0.14.mdx
benoitgrelard: It's not. Looks confusing but it reads like this:

image

That's what was there before, I assumed whoever put it there had good reasons 😀

PR #183, opened 2021-06-15
[DropdownMenu] Update docs

andy-hook:


data/primitives/components/dropdown-menu/0.0.20.mdx
```diff
@@ -490,6 +512,46 @@ The component that contains the dropdown menu items.
   ]}
 />
 
+### TriggerItem
+
+An item that opens a submenu. Used in combination with a nested `DropdownMenu`. Must be rendered inside `DropdownMenu.Root`.
```
benoitgrelard, line 517: That seems odd, it shouldn't be in the root right?

benoitgrelard (commented):
- We should update the anatomy section too to reflect show how nested menus appear.
- I think we need to the same type changes for `onPointerDownOutside`/`onFocusOutside`/`onInteractOutside` as I've done in #180 for example
- Can we add a slight negative `alignOffset` on submenus to make sure parent/child items are aligned like in macOS?

image

data/primitives/components/dropdown-menu/0.0.20.mdx
andy-hook: I think that's right, it's basically in place of a `Trigger` which is rendered directly inside of `root`, usually as the first child right?

benoitgrelard: Ah yes of course, my bad!

benoitgrelard (approved):

PR #193, opened 2021-06-23
Remove extra angle bracket from polymorphic primitive basic example

brattonross:
Removes an extra angle bracket from the polymorphic primitive basic example, which causes a typescript error when copy-pasted directly into an editor.

This pull request:

- [ ] Fixes a bug
- [ ] Other


peduarte (approved):

peduarte:
Thanks! :D

PR #218, opened 2021-08-03
[Toast] New primitive docs

jjenzz:
This still needs a demo but covers all the other areas. It'd be good to see if @peduarte is able to create a demo from following these docs when he is back. 

I've added more detail than we usually do in our accessibility and custom API sections.

### DIFF
--- a/components/demos/Toast.jsx
+++ b/components/demos/Toast.jsx
@@ -87,13 +87,13 @@ const StyledAction = styled(ToastPrimitive.Action, {
 });
 
 // Exports
-const ToastProvider = ToastPrimitive.Provider;
-const ToastViewport = StyledViewport;
-const Toast = StyledToast;
-const ToastTitle = StyledTitle;
-const ToastDescription = StyledDescription;
-const ToastAction = StyledAction;
-const ToastClose = ToastPrimitive.Close;
+export const ToastProvider = ToastPrimitive.Provider;
+export const ToastViewport = StyledViewport;
+export const Toast = StyledToast;
+export const ToastTitle = StyledTitle;
+export const ToastDescription = StyledDescription;
+export const ToastAction = StyledAction;
+export const ToastClose = ToastPrimitive.Close;
 
 // Your app...
 const Box = styled('div', {});
--- a/components/demos/Toast.jsx
+++ b/components/demos/Toast.jsx
@@ -1,0 +1,196 @@
+import * as React from 'react';
+import { styled, keyframes } from '@modulz/design-system';
+import { violet, blackA, mauve, slate, green } from '@radix-ui/colors';
+import * as ToastPrimitive from '@radix-ui/react-toast';
+
+const VIEWPORT_PADDING = 25;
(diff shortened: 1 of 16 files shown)


components/PropsTable.tsx
```diff
@@ -78,7 +78,13 @@ export function PropsTable({
...
                     </AccessibleIcon>
                   </PopoverTrigger>
-                  <PopoverContent side="top">
+                  <PopoverContent
+                    side="top"
+                    onOpenAutoFocus={(event) => {
+                      event.preventDefault();
+                      (event.currentTarget as HTMLElement)?.focus();
+                    }}
+                  >
```
jjenzz, line 87: This popover in the `PropsTable` is used to describe the props so may contain links. I've prevented the link from becoming focused and instead focused the content here as it's unlikely these would contain form controls and things.

peduarte:
Nice one. Yeah I can create a demo for this 🚀

data/primitives/components/toast/0.1.0.mdx
```diff
@@ -0,0 +1,573 @@
...
+---
+
+# Toast
+
+<Description>A succinct message that is displayed temporarily.</Description>
+
+<HeroContainer>
+  <ToastDemo />
+</HeroContainer>
+
```
benoitgrelard, line 15: I think you're missing the bit that shows the code and the codesanbox link

components/HeroContainer.tsx
```diff
@@ -16,6 +16,7 @@ export function HeroContainer({ css, children }: { css?: any; children?: React.R
         justifyContent: 'center',
         py: 100,
         borderRadius: '$3',
+        overflow: 'hidden',
```
andy-hook, line 19: I'd be careful here as not all demos will want to clip this. Perhaps better to handle this inline at the call site?

jjenzz: Ah oops! I didn't mean to leave this in. I added it when I had the toast sliding in from the side of the demo but forgot to remove it... good catch 👍

components/demos/Toast.jsx
```diff
@@ -0,0 +1,196 @@
...
+});
+
+// Exports
+const ToastProvider = ToastPrimitive.Provider;
+const ToastViewport = StyledViewport;
+const Toast = StyledToast;
+const ToastTitle = StyledTitle;
+const ToastDescription = StyledDescription;
+const ToastAction = StyledAction;
+const ToastClose = ToastPrimitive.Close;
```
benoitgrelard, line 96: I think these are usually exported.

andy-hook: I was wondering about why that is?

benoitgrelard: It's just to simulate what a user might do really, showing that these would be exported from their DS/component file.

jjenzz: I'll update the other demos as well then as part of this because I coped this from `Dialog`.

PR #219, opened 2021-08-03
Primitives release 04/08/2021 – Setup, changelog

andy-hook:
First PR of a series which are going into a `release-04-08-2021` branch for docs changes.

> Note: build will fail until packages have been published

- [x] Publish release
- [x] Update `lock` file


data/primitives/overview/releases.mdx
```diff
@@ -7,6 +7,64 @@ metaDescription: Radix Primitives releases and their changelogs.
...
+<PackageRelease name="Dialog" version="0.0.20" />
+
+- ✨ Add Modality support via `modal` prop <PRLink id={700} />
+- 🔥 Remove `Dialog.Content` `disableOutsideScroll` prop <PRLink id={700} />
+- 🔥 Remove `Dialog.Content` `disableOutsidePointerEvents` prop <PRLink id={700} />
+- 🔥 Remove `Dialog.Content` `trapFocus` prop <PRLink id={700} />
+- 🐛 Improve animation rendering in React 18 <PRLink id={776} />
+- 🐛 Ensure focus is restored to trigger on close when using the `autofocus` attribute on a child element <PRLink id={739} />
+- 🐛 Prevent outside pointer events triggering prematurely on touch devices <PRLink id={767} />
+- 🐛 Ensure inputs are consistently focused after open <PRLink id={776} />
```
jjenzz, line 38:
```suggestion
- 🐛 Ensure first focusable element consistently receives focus in iOS Safari <PRLink id={776} />
```

I was about to suggest this as a missing item but I think this last one is the same thing? Can we replace the other places this item has been added too if that's the case 🙏

jjenzz (commented):
Looks good to me, just one small request 👍

data/primitives/overview/releases.mdx
andy-hook: Ah yes this is much better 👍

data/primitives/overview/releases.mdx
```diff
@@ -7,6 +7,58 @@ metaDescription: Radix Primitives releases and their changelogs.
 
 <Description>Radix Primitives releases and their changelogs.</Description>
 
+## 04/08/2021
+
+**All Primitives**
+
+- ✨ Improve polymorphic type performance <PRLink id={784} />
```
jjenzz, line 14: I wonder if we should do this given the bug report?

```suggestion
- 🔥 Improve polymorphic type performance <PRLink id={784} />
```

PR #229, opened 2021-08-25
Add new 0.1.0 pages

peduarte:
This PR just duplicates the files without making any changes to them yet.
The changes will come in the next PR.

> Note: This adds types for the routes and sets `Slot` and `Polymorphic` packages as deprecated.


peduarte:
@benoitgrelard but we're moving Radix to Beta, so all components should start at `0.1.0`, right

benoitgrelard:
> @benoitgrelard but we're moving Radix to Beta, so all components should start at `0.1.0`, right?

Yes, you're right sorry about the confusion. That's what we've done on Friday, we've bumped them all.

PR #233, opened 2021-09-06
Deprecate extendPrimitive

benoitgrelard:


data/primitives/overview/styling.mdx
```diff
@@ -131,29 +131,20 @@ export default () => {
 
 ## Extending a primitive
 
-We expose an `extendPrimitive` utility as a convenience for adding default props. It can be useful if your preference is to style data attributes as they needn't be composed like class names.
+Extending a primitive is done the same way you extend any React component. One minor difference is the use of `Radix.ComponentPropsWithoutRef` instead of `React.ComponentPropsWithoutRef` which is just a workaround until [the fixed types are released](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/55396).
```
jjenzz, line 134:
```suggestion
Extending a primitive is done the same way you extend any React component. One minor difference is the use of `Radix.ComponentPropsWithoutRef` instead of `React.ComponentPropsWithoutRef` while we await review of a [`@types/react` update](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/55396).
```

the main change I'm trying to propose is to remove  "the fixed types" part because there is a chance there is a reason the types are the way they are on DefinitelyTyped and my PR may get rejected.

It might not be likely but probably better off not making assumptions in our wording.

jjenzz (changes requested):

data/primitives/overview/styling.mdx
jjenzz: Oh, having said all that I've just seen that they've [gotten around to reviewing it now](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/55396#issuecomment-911796016) and sounds like it might be a goer. So, feel free to ignore this!

benoitgrelard: Yeah I checked before and it seems they agree on the correctness of your fix, so I thought I'd mention here.

peduarte (approved):
LGTM

peduarte (approved):

PR #240, opened 2021-09-16
[Checkbox][Slider][Switch][RadioGroup] Add information about rendered `input`

jjenzz:
Someone that asked me for help showed me examples where they've wrapped these components to add their own `input` for use in forms and I had to explain that we already do that for them. 

I'm not sure how we want to communicate this so have added something to discuss.

- Updated examples with `form` so that viewing source shows `input`
- Added copy to explain that we do this for them when used in a `form`.


data/primitives/components/checkbox/0.1.0.mdx
```diff
@@ -53,7 +53,7 @@ export default () => (
 
 ### Root
 
-Contains all the parts of a checkbox.
+Contains all the parts of a checkbox. It will render an `input` when used within a `form` to ensure events propagate correctly.
```
benoitgrelard, line 56: Wondering if we should word it a bit differently cause this makes it sound like that part will actually render as an input when it's rather that an extra input node gets added internally right?

benoitgrelard (commented):
I don't have anything against these changes, makes sense to me.
Just left one minor comment.

data/primitives/components/checkbox/0.1.0.mdx
jjenzz: Changed the copy but not sure if it's better... what do you think?

benoitgrelard: I think it's better yeah.

benoitgrelard (approved):

PR #243, opened 2021-09-29
[very minor] Added a title to the webpage

tnb24:
I opened the [website](https://www.radix-ui.com/) and released it didn't have a title so, I added it. 

I also removed the radix UI logo import as it was unused and nested all my elements in a ```<> & </>``` as JSX elements need to have a parent element,

It's a pretty minor so I've tested it 'thoroughly'. Thanks for making this open source :smile:

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import NextLink from 'next/link';
+import Head from 'next/head';
 import {
   Section,
   Link,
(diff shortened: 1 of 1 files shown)


peduarte:
Hey. This is great, thank you so much. I appreciate it a lot.

However we're in the process of re-doing the home page, and this is being addressed. So I'll close to avoid unnecessary conflicts.

Thank you, though 🙏

PR #250, opened 2021-10-11
Update animation example to use `asChild`

andy-hook:
https://github.com/radix-ui/primitives/issues/904


andy-hook:
> are the other changes updating for a newer version of `react-spring`?

Yup, seems this [changed in v9](https://aleclarson.github.io/react-spring/v9/breaking-changes/#The-transition-function) so updated to match.

PR #252, opened 2021-10-13
Update DS

peduarte:
From: https://github.com/radix-ui/design-system/pull/357


jjenzz:
I updated the lockfile here otherwise i was seeing these warnings locally:

image

A quick test of the app seems okay.

Issue #265, opened 2021-10-28
Blurry Alert Dialog in documentation


rowbot-weisguy:
### Relevant Radix Component(s)
Alert Dialog

### Recommendation
Hello awesome radix team!

I noticed that in Chrome 94 the Alert Dialog doc example looks blurry when opened. I've noticed that sometimes css `transform` can cause elements to lose their sharpness, but in particular I noticed it's the `will-change: 'transform'` that causes the blurriness in this example. 

Here's the original CodeSandbox, shown on the docs page: https://codesandbox.io/s/nmk9d?module=App.js
Here's my fork of the CodeSandbox: https://codesandbox.io/s/divine-fog-lznb2?module=App.js

It might be wise to remove the `will-change` so that the demo component looks awesome in all browsers. It would be a shame if somebody thought it was the primitive's fault. 

Cheers!

PR #266, opened 2021-10-29
[AlertDialog][Dialog] Remove will-change from demos

andy-hook:
closes https://github.com/radix-ui/website/issues/265

Unsure if this was originally needed for perf but it does improve the clarity quite a bit.


benoitgrelard (approved):
Works for me.

Worth thinking about potentially updating those default styles to a different alignment technique too once we release the changes for the portal part.

PR #276, opened 2021-11-16
Fix sandbox type error

andy-hook:
We had reports of type errors in our demos
image

I'm still investigating but nothing changed on our end AFAIK, for now I've found adding `react-scripts` as a dependency to the generated sandbox resolves the issue.


andy-hook:
This was confirmed as an issue by the CSB team and they are working on a fix (~1 day). I went ahead and merged this as a workaround in the meantime.

PR #278, opened 2021-11-17
[Sandbox demos] Move react-scripts to dev dependencies

andy-hook:
Original fix https://github.com/radix-ui/website/pull/276 should have added this as `devDependencies`

CSB came back with the same solution

> The sandbox didn't have react-scripts installed, which made our detection logic think that you were on react-scripts v1, which imports .cjs files as raw files. If you add react-scripts v4 to your devDependencies, we'll use the latest version of react-scripts, and it will work.

When asking about the reason this requirement would have changed:
> We did recently change our resolver logic, and we resolved the .cjs file instead of the .mjs file. That's the only thing that could explain it (I think).

### DIFF
--- a/components/DocCodeBlock.tsx
+++ b/components/DocCodeBlock.tsx
@@ -168,11 +168,13 @@ const makeCodeSandboxParams = (name, code) => {
           dependencies: {
             react: 'latest',
             'react-dom': 'latest',
-            'react-scripts': 'latest',
             '@stitches/react': 'latest',
             '@radix-ui/colors': 'latest',
             '@radix-ui/react-icons': 'latest',
             [`@radix-ui/react-${name}`]: 'latest',
+          },
+          devDependencies: {
+            'react-scripts': 'latest',
           },
         } as any,
         isBinary: false,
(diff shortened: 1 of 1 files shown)


components/DocCodeBlock.tsx
```diff
@@ -196,7 +198,6 @@ ReactDOM.render(<div><App /></div>, document.getElementById('root'));`,
         isBinary: false,
       },
     },
-    template: 'create-react-app',
```
andy-hook, line 199: This is [automatically detected](https://github.com/codesandbox/codesandbox-importers/blob/cd59770d734da1b601e7c461218bee1dff86c825/packages/import-utils/src/create-sandbox/templates.ts#L163-L165) given the presence of `react-scripts`. Though could argue it's safer to just leave it explicitly specified?

peduarte:
Trying it out in the [preview build](https://radix-website-git-update-sandbox-deps.modulz-deploys.com/docs/primitives/components/accordion) but im getting an error 
CleanShot 2021-11-22 at 12 00 11@2x

andy-hook:
> Trying it out in the [preview build](https://radix-website-git-update-sandbox-deps.modulz-deploys.com/docs/primitives/components/accordion) but im getting an error CleanShot 2021-11-22 at 12 00 11@2x

@peduarte Are you still seeing this? not able to reproduce on my side.

peduarte:
Looks good now 🚀

PR #275, opened 2021-11-15
Hero carousel improvements

vladmoroz:
- Make the carousel draggable
- Add dialog demo (thanks @jjenzz for that portal change!)
- Add “See more components in the docs” slide
- Scroll slides into view more consistently when roving through the carousel with keyboard arrows
- Update Radix stats (@colmtuite not sure about downloads, what's the latest number?)

PR #279, opened 2021-11-17
[Dialog] Revert sandbox demo code

andy-hook:
Looks like our demo code for dialog was updated in https://github.com/radix-ui/website/pull/275 so that we can use the new `Portal` part. This [caused issues with the generated sandbox](https://codesandbox.io/s/ds0g9?module=App.js) because all the radix deps get pegged `latest`.

This change reverts to the previous code and uses an alias for the RC version so that the carousel can still make use of the features. Once the new api is available on `latest` we can remove this fix.

@vladmoroz I think this works ok, let me know if anything looks off.


jjenzz (approved):
Seems sensible to me 👍

vladmoroz (approved):
Oh, thanks for catching this Andy.
I was worried about breaking something as I was updating the dialog dependency, but didn't know to check CodeSandbox.

PR #294, opened 2021-12-09
[Release] Remove live demos from old versions

jjenzz:


jjenzz:
I've made this part of the other PR because it makes more sense there.

PR #295, opened 2021-12-09
[Release] Update docs to reflect release changes

jjenzz:


components/demos/Tooltip.jsx
```diff
@@ -73,17 +74,30 @@ const IconButton = styled('button', {
...
+        <TooltipTrigger asChild css={{ marginLeft: 5 }}>
+          <IconButton>
+            <MinusIcon />
+          </IconButton>
+        </TooltipTrigger>
+        <StyledContent sideOffset={5} className={`${theme}`}>
+          Remove from library
+          <StyledArrow />
+        </StyledContent>
+      </Tooltip>
```
jjenzz, line 99: I added another tooltip here to demonstrate the benefit of the provider. Thoughts?

data/primitives/components/alert-dialog/0.1.1.mdx
```diff
@@ -88,6 +88,31 @@ Contains all the parts of an alert dialog.
...
+          . See <Link
+            variant="blue"
+            href="https://github.com/theKashey/react-remove-scroll#usage"
+            target="_blank"
+          >
+            their docs
+          </Link> for information on functionality and limitations.
+        </span>
+      ),
+    },
```
jjenzz, line 115: This existed in `0.1.1` but forgot to add it there by the looks of it.

components/MDXComponents.tsx
```diff
@@ -239,9 +238,5 @@ export const FrontmatterContext = React.createContext<Frontmatter>({} as any);
 // https://github.com/hashicorp/next-mdx-remote#using-providers
 export function MDXProvider(props) {
   const { frontmatter, children } = props;
-  return (
-    <IdProvider>
```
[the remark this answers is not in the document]
jjenzz: @peduarte Can you help with this please? 🙏

data/primitives/overview/server-side-rendering.mdx
```diff
@@ -16,22 +16,10 @@ Server side rendering or `SSR`, is a technique used to render components to HTML
...
-```tsx
-import { IdProvider } from '@radix-ui/react-id';
-
-export default () => (
-  <IdProvider>
-    <App />
-  </IdProvider>
-);
-```
+In other words, the equivalent of [Time to Interactive](https://web.dev/interactive/) for screen reader users will depend on the download speed of the JS bundle, unless using React 18.
```
benoitgrelard, line 25: Not sure how to reword, but it would be good to explain that whilst things generally work with 17 in a non-optimal way, if you're after SSR, the better path forward is probably to adopt 18.

lib/primitivesRoutes.ts
```diff
@@ -51,10 +51,10 @@ export const primitivesRoutes = [
     pages: [
       { title: 'Accessible Icon', slug: 'docs/primitives/utilities/accessible-icon' },
       { title: 'Announce', slug: 'docs/primitives/utilities/announce' },
-      { title: 'Id Provider', slug: 'docs/primitives/utilities/id-provider' },
+      { title: 'Id Provider', slug: 'docs/primitives/utilities/id-provider', deprecated: true },
```
benoitgrelard, line 54: Would be good to add a deprecation notice on the page itself like we had somewhere else, and then linked back to the SSR page to explain the new path forward.

components/MDXComponents.tsx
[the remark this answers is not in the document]
jjenzz: @benoitgrelard I can't actually find any client/server mismatch errors in the codebase so I dunno if we need to upgrade the DS. In fact, even if we did get those errors they wouldn't break anything?

benoitgrelard (approved):

components/MDXComponents.tsx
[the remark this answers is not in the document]
benoitgrelard: They wouldn't, but we usually make sure the DS dep on primitives are aligned with these on the site, otherwise it means there are 2 versions of each on the site and some potential weird situations.

components/demos/Tooltip.jsx
peduarte: Love to see the `Provider` getting shipped! So useful. 

I 100% think we need an example for it, but I'm thinking it should be under the _Examples_ section, without a live playground, because generally the hero example is kept simpler

Are you guys cool with that?

components/MDXComponents.tsx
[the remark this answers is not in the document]
peduarte: Not sure what's happening here but happy to chat about it on Slack 🚀

benoitgrelard (approved):

PR #310, opened 2021-12-13
Add information about pinning

jjenzz:
Add recommendation on pinning to getting started documentation.


peduarte (approved):
I'd have `beta` as lowercase personally, but all good, looks great!

data/primitives/overview/getting-started.mdx
```diff
@@ -15,10 +15,10 @@ In this quick tutorial, we will install, style, and export a [Popover](../compon
 
 ### 1. Install the primitive
 
-Install the component from your command line.
+Install the component from your command line. We recommended pinning to a specific version to avoid breaking changes while we are in Beta.
 
 ```bash
-npm install @radix-ui/react-popover
+npm install @radix-ui/react-popover@0.1.3
```
benoitgrelard, line 21: I know there's a way with yarn to install the latest whilst pinned, which I think would be a better way to tell people to install rather than put a version here.

I think it's `--exact` but not sure if `npm` also has the same thing or not.

peduarte, line 21: just thinking about the maintenance of this... are we cool we updating this guide every time theres a new popover release?
also, I dont think this pins the version, it just installs that specific version but keeps the `^`

we can use the `-E/--exact` flag to pin

thinking we could do something like this:
```sh
npm install @radix-ui/react-popover@latest -E
```

PR #312, opened 2021-12-16
[@radix-ui/react-dropdown-menu] lack of "dir" prop in Content

Luccasoli:
- [X] Test the change in your own code.

This pull request:

- [ ] Fixes a bug
- [ ] Other

### DIFF
--- a/data/primitives/components/dropdown-menu/0.1.4.mdx
+++ b/data/primitives/components/dropdown-menu/0.1.4.mdx
@@ -379,6 +379,18 @@ The component that pops out when the dropdown menu is open.
         </span>
       ),
     },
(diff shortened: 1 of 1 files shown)


jjenzz:
@Luccasoli Thanks for contributing. The `dir` "prop" is a HTML attribute though.

We only document custom props because all of our components support applicable HTML attributes which would be a lot to document as individual line items. I've been meaning to update the docs to clear this up as it isn't obvious that we support these (or all of React's built in Synthetic Events) unless using TypeScript.

I'll close this for now and create a Github Issue to remind us to make this all a bit clearer because the solution described above would require some design input.

PR #314, opened 2021-12-20
Update Releases page design

vladmoroz:


colmtuite:
Thanks for sorting this PR, it's a big improvement imo. The square bracket wins for me, but I don't have a strong preference over option 3. Option 2 doesn't work imo. Perhaps a group vote? 😄

jjenzz:
Also, on the discussion about the page being too long, have we considered an accordion? I fiddled with this the other evening and quite liked it. https://codesandbox.io/s/objective-fast-8kov6?file=/src/App.js

vladmoroz:
Sure, popular vote be it.

@jjenzz I like how succinct the accordion looks and feels, but then you can't ⌘F random stuff within accordion items' content 😕. Personally, I rely on this native browser search a lot when skimming developer docs and don't want to give it up.

***

@colmtuite do you want to prepare page per release in this iteration? Should be pretty easy. I'm thinking that on the /releases pages, nothing changes except that the release headings link to a separate page with content just for that release. How does that sound?

jjenzz:
> but then you can't ⌘F random stuff within accordion items' content

Fair, but wouldn't you be doing `⌘F` for the component name? How would you know what copy the release notes contain in advance in order to browser search it?

Although tbh, if we're eventually doing release notes per page then the accordion isn't really worth it in the meantime.

colmtuite:
@vladmoroz I'd say let's ship this portion, then loop back around to split it into individual pages. Mostly because I don't have much time this week, and it would be nice to have some time to explore a visual timeline for when we move to individual release pages.

colmtuite (approved):

data/primitives/overview/releases.mdx
```diff
@@ -7,561 +7,595 @@ metaDescription: Radix Primitives releases and their changelogs.
...
+- Improve React 18 support <PRLink id={984} />
+- Improve dev mode errors with mismatched `type` and `value` props <PRLink id={979} />
+- Prevent `Accordion.Content` height animation on initial page load <PRLink id={977} />
 
-- 🔥 Add new `Portal` part that allows portalling into custom container <PRLink id={936} />
-- 🔥 Support scrolling within `AlertDialog.Overlay`. Move `allowPinchZoom` to root. <PRLink id={963} />
-- 🐛 Fix `asChild` TypeScript error <PRLink id={924} />
+<PackageRelease name="Alert Dialog" version="0.1.5" major />
+
+- **Breaking** – Add new `Portal` part that allows portalling into custom container <PRLink id={936} />
```
jjenzz, line 26: Same here, copy could be clearer:

> Add new `Portal` part. To avoid regressions, use this part if you want portalling behavior. 

This copy is used elsewhere in the release too so would be good to copy it there too if we decide to tweak.

data/primitives/overview/releases.mdx
```diff
@@ -7,561 +7,595 @@ metaDescription: Radix Primitives releases and their changelogs.
...
+## December 13, 2021
 
-**All Primitives**
+This release focuses on React 18 support and introduces a number breaking changes to some packages, mostly related to portalling dialogs.
 
-- 🔥 Remove `IdProvider` and improve React 18 support for generated ids <PRLink id={1006} />
+<PackageRelease name="All primitives" />
 
-<PackageRelease name="Accordion" version="0.1.5" />
+- **Breaking** – Remove `IdProvider` and improve React 18 support for generated ids <PRLink id={1006} />
```
jjenzz, line 16: Maybe make this one clearer too?

> Deprecate `IdProvider`. Improves support for React 18 going forward and is no longer needed in older versions. Remove from your app to avoid deprecation warnings.

jjenzz (commented):
Can also do these in a separate PR if we want but think it would be useful to tweak the copy for these given the discussions on Slack about the copy not being clear.

jjenzz (approved):

vladmoroz:
Thanks @jjenzz for the content suggestions (3590dd5).

Merging!

PR #317, opened 2021-12-28
[website] Prevent header shifting while scrolllocking

PookMook:
- [X] Test the change in your own code.
^ Vercel bot will be a good soldier and add it for us

This pull request:

- [X] Fixes a bug
- [ ] Other

=========================================

react-remove-scroll expects you to add some classes to the containers potentially pushed around by the "removal" of the scrollbar, this PR adds it to the header (the body and quick nav already have what they need to not shift around)

=========================================

## Current behavior
without2
without

## With this change applied
with2
withClass

### DIFF
--- a/components/Header.tsx
+++ b/components/Header.tsx
@@ -21,13 +21,14 @@ import { PlusIcon, MixIcon, StitchesLogoIcon } from '@radix-ui/react-icons';
 import { ThemeToggle } from '@components/ThemeToggle';
 import { BoxLink } from '@components/BoxLink';
 import { RadixLogoIcon } from './RadixLogoIcon';
+import { RemoveScroll } from 'react-remove-scroll';
 
 export const Header = () => {
   const router = useRouter();
   const isColors = router.pathname.includes('/colors') || router.pathname.includes('/docs/colors');
 
   return (
-    <Box as="header" className="width-before-scroll-bar">
+    <Box as="header" className={RemoveScroll.classNames.fullWidth}>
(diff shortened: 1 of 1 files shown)


components/Header.tsx
```diff
@@ -27,7 +27,7 @@ export const Header = () => {
   const isColors = router.pathname.includes('/colors') || router.pathname.includes('/docs/colors');
 
   return (
-    <Box as="header">
+    <Box as="header" className="width-before-scroll-bar">
```
benoitgrelard, line 30: I think a constant can be imported from the package we use.
I believe we do this already somewhere.
I think it'd be something like

```
import { RemoveScroll } from 'react-remove-scroll';

…

className={RemoveScroll.classNames.fullWidth}
```

Reference: https://github.com/theKashey/react-remove-scroll-bar/blob/master/src/constants.ts

benoitgrelard (commented):
Thanks for the contribution!

components/Header.tsx
PookMook: That's a great point, wasn't aware react-remove-scroll provided those constants
Made the change in https://github.com/radix-ui/website/pull/317/commits/c9c031ed1385f454eab6855f77a8caae08a64af3

benoitgrelard (approved):

components/Header.tsx
benoitgrelard: Amazing, Thanks!

PR #322, opened 2022-01-10
[ToggleGroup] Preventing undefined example

jjenzz:
I feel like I've answered this question a few times now so adding an example to the docs

### DIFF
--- a/data/primitives/components/toggle-group/0.1.4.mdx
+++ b/data/primitives/components/toggle-group/0.1.4.mdx
@@ -244,7 +244,7 @@ An item in the group.
 
 ## Examples
 
-### Prevent undefined value
+### Ensuring there is always a value
 
 You can control the component to ensure a value.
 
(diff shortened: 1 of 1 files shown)


benoitgrelard (approved):
All good, the only thing I'd consider is maybe naming the example a bit differently as I'm not sure this is how people think of it. I think most people would say something like: "I want it so that you can't unselect", "I want it so there's always a value", not sure what would be best but I myself had to double check what it was about after reading the example name.

It kinda feels like a double negative, so maybe instead of prevent undefined, it's more like, ensure there is a value?

jjenzz:
> It kinda feels like a double negative, so maybe instead of prevent undefined, it's more like, ensure there is a value?

@benoitgrelard Haha I'm glad you raised this because I was really struggling with the naming here. I tried:

- Minimum value
- Prevent deselect
- At least one value
- Keep one pressed

Started to go a bit crazy. Do you have any specific ideas?

benoitgrelard (approved):

PR #325, opened 2022-01-26
“When to use” section

vladmoroz:
For now, check Hover Card and Popover pages – if we agree on the direction here, will add content for all or most components

### DIFF
--- a/components/MDXComponents.tsx
+++ b/components/MDXComponents.tsx
@@ -199,6 +199,28 @@ export const components = {
   PRLink,
   HeroContainer,
   HeroQuote,
+  Accordion: DS.Accordion,
+  AccordionItem: DS.AccordionItem,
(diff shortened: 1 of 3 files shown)


colmtuite:
The section heading confused me for some reason. Perhaps "Frequently asked questions" might be better? It would be a more common heading for this type of info imo.

In the second DropdownMenu answer, it mentions "toggle button", which confused me, because I was thinking of our ToggleGroup component. Just "button" or "trigger" might be more apt?

benoitgrelard:
> In the second DropdownMenu answer, it mentions "toggle button", which confused me, because I was thinking of our ToggleGroup component. Just "button" or "trigger" might be more apt?

Yeah agreed, we should try and keep consistent with the language we use overall, so these would be "trigger".

PR #332, opened 2022-02-18
[Select] New primitive docs

benoitgrelard:
- [x] Add demo
- [x] Flesh out some examples

### DIFF
--- a/data/primitives/components/select/0.1.0.mdx
+++ b/data/primitives/components/select/0.1.0.mdx
@@ -635,6 +635,8 @@ export default () => (
 
 Adheres to the [ListBox WAI-ARIA design pattern](https://www.w3.org/TR/wai-aria-practices-1.2/#Listbox).
 
+See the W3C [Select-Only Combobox](https://www.w3.org/TR/wai-aria-practices/examples/combobox/combobox-select-only.html) example for more information.
+
 ### Keyboard Interactions
 
 <KeyboardTable
--- a/data/primitives/components/select/0.1.0.mdx
+++ b/data/primitives/components/select/0.1.0.mdx
@@ -585,7 +585,7 @@ export default () => {
 
 The native scrollbar is hidden by default as we recommend using the `ScrollUpButton` and `ScrollDownButton` parts for the best UX. If you do not want to use these parts, compose your select with our [Scroll Area](scroll-area) primitive.
 
-```jsx line=5-8,10-13,15,17-20,27-28,32-36
+```jsx line=5-8,10-13,15-18,20-23,29,31,35,37-40
 import { styled } from '@stitches/react';
 import * as Select from '@radix-ui/react-select';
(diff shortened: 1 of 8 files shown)


components/PrimitivesPage.tsx
```diff
@@ -108,6 +109,11 @@ export function PrimitivesPage({ children }: { children: React.ReactNode }) {
                       <Text size="2" css={{ color: 'inherit', lineHeight: '1' }}>
                         {page.title}
                       </Text>
+                      {isBeta ? (
+                        <Badge css={{ ml: '$2' }} variant="blue">
+                          Beta
+                        </Badge>
+                      ) : null}
```
jjenzz, line 116: mind if we abstract this change into its own PR so that andy and i can update our prs with the same?

data/primitives/components/select/0.1.0.mdx
```diff
@@ -0,0 +1,765 @@
...
+    </Select.Content>
+  </Select.Root>
+);
+```
+
+### Controlling the value displayed in the trigger
+
+By default the trigger will automatically display the selected item `ItemText`'s content. You can control what appears by chosing to put things inside/outside the `ItemText` part.
+
+If you need more flexibility, you can controlling the component using `value`/`onValueChange` props and passing `children` to `SelectValue`. Remember to make sure what you put in there is accessible.
```
jjenzz, line 548:
```suggestion
If you need more flexibility, you can control the component using `value`/`onValueChange` props and passing `children` to `SelectValue`. Remember to make sure what you put in there is accessible.
```

components/PrimitivesPage.tsx
benoitgrelard: Sure I'll do that now.

data/primitives/components/select/0.1.0.mdx
```diff
@@ -0,0 +1,765 @@
...
+        </ScrollAreaScrollbar>
+      </ScrollArea>
+    </Select.Content>
+  </Select.Root>
+);
+```
+
+## Accessibility
+
+Adheres to the [ListBox WAI-ARIA design pattern](https://www.w3.org/TR/wai-aria-practices-1.2/#Listbox).
```
jjenzz, line 636: i wonder if it it is worth mentioning it follows the example here too? https://www.w3.org/TR/wai-aria-practices/examples/combobox/combobox-select-only.html

i know we haven't done that elsewhere but i quite liked that andy did it in his 😅

components/PrimitivesPage.tsx
benoitgrelard: #341

data/primitives/components/select/0.1.0.mdx
benoitgrelard: I'll take a look closer at what he did, happy to add more details.

jjenzz (approved):

PR #340, opened 2022-02-22
[Release] Document `Dialog` labelling changes

benoitgrelard:


data/primitives/components/dialog/0.1.6.mdx
```diff
@@ -314,7 +314,9 @@ The button that closes the dialog.
 
 ### Title
 
-An accessible name to be announced when the dialog is opened. Alternatively, you can provide `aria-label` or `aria-labelledby` to `Dialog.Content` and exclude this component.
+An accessible title to be announced when the dialog is opened.
+
+If you don't want the title to be visible, wrap it inside `<VisuallyHidden asChild>`. Alternatively, you can provide `aria-label` or `aria-labelledby` to `Dialog.Content` and exclude this component.
```
andy-hook, line 319: Should we link the mention of `VisuallyHidden` to its docs page to make it clearer that it's an additional package?

jjenzz: they can't just add `aria-label` anymore because it won't silence the `Title` warning. the `Title` part is required now.

benoitgrelard: I'm confused, this seems to suggest otherwise:

image

jjenzz: yeah, the `aria-describedby={undefined}` was just to remove the `id` that we add in cases where the `Description` isn't rendered 😢  

we struggled to come up with a way around that without it breaking things for SSR.

benoitgrelard: Yeah I remember, that's why we landed on that, but it's a shame we can't abstract `aria` there then. To be fair we should encourage people to use descriptions anyway.

jjenzz (approved):

PR #353, opened 2022-03-10
`id` prop belongs to DropdownMenu.Content instead of Root

### LINKED ISSUE #356: Id functionality and documentation for DropdownMenu.Root
## Bug report
Using `id` is not working as expected(as detailed in the documentation) and types do not contain `id`. 

### Current Behavior
Using `id` is not working as expected(as detailed in the documentation) and types do not contain `id`. 

### Expected behavior
Types to include `id` and to pass id attr to ref element, and cases `asChild` also to be contemplated.

### Reproducible example
Reproduced with the latest version "@radix-ui/react-dropdown-menu": "^0.1.6",

### Additional context
Docs:
Screen Shot 2022-03-10 at 11 25 43

Types:
Screen Shot 2022-03-10 at 11 25 32

Repro: Not passing the right `id` (using asChild true)

Screen Shot 2022-03-10 at 11 29 25

### Your environment

| Software         | Name(s) | Version |
| ---------------- | ------- | ------- |
| Radix Package(s) |         |        0.1.6 |
| React            | n/a     |    18     |
| Browser          |    Chrome     |        |
| Assistive tech   |         |         |
| Node             | n/a     |         |
| npm/yarn         |         |         |
| Operating System |         |         |

okbel:
As detailed [here](https://github.com/radix-ui/primitives/issues/1238), passing `id` to .Root doesn't work as expected (and as detailed in the documentation).
Moving this to .Content

This pull request:
- [ ] Fixes a bug
- [ ] Other

Closes #356

### DIFF
--- a/data/primitives/components/dropdown-menu/0.1.6.mdx
+++ b/data/primitives/components/dropdown-menu/0.1.6.mdx
@@ -147,17 +147,6 @@ Contains all the parts of a dropdown menu.
       description:
         'The reading direction of submenus when applicable. If omitted, assumes LTR (left-to-right) reading mode.',
     },
-    {
-      name: 'id',
-      type: 'string',
-      description: (
-        <span>
-          A unique identifier for the component. The <Code>id</Code> is used to
-          generate <Code>id</Code> attributes for nested components. If no{' '}
-          <Code>id</Code> prop is provided, a generated id will be used.
-        </span>
-      ),
-    },
   ]}
 />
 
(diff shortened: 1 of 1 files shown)


benoitgrelard (changes requested):
Yep that looks like it was a mistake from us.

However, we don't generally list DOM props like these on parts as they are implied by the element we render.

Would you mind just deleting it, rather than moving it?

Thanks!

benoitgrelard:
I've handled it in #365

PR #354, opened 2022-03-24
fix aria-labels on Carousel Previous/Next buttons

jonmilner:
👋 hi! big fan of Radix and was poking around the source code, just noticed that the aria labels are swapped on the home page carousel previous/next buttons and figured I'd open a quick PR in case no one's caught it yet

- [ x] Use a meaningful title for the pull request. Include the name of the package modified.

This pull request:

- [ x] Fixes a bug
- [ ] Other


benoitgrelard (approved):
Thanks for that!

PR #377, opened 2022-04-28
Dialog: Properly implement custom ref

pfcodes:
Correctly implements `ref`. Setting useState worked, but it was triggering errors in Typescript do to `setContainer` not being a valid `ref` prop.

This pull request:

- [ ] Fixes a bug
- [ ] Other


jjenzz:
this was correct the way it was i believe, so i think the solution would be to solve the type issue.

in React, the `div` the ref is attached to will render **after** the portal has rendered, so the container prop will be `null` and the ref assignment will not trigger a re-render to update it. 

this issue is more apparent when the dialog is open by default.

PR #385, opened 2022-05-05
Document `NavigationMenu.Link` improvements

andy-hook:
https://github.com/radix-ui/primitives/pull/1372

### DIFF
--- a/data/primitives/components/navigation-menu/0.1.3.mdx
+++ b/data/primitives/components/navigation-menu/0.1.3.mdx
@@ -345,6 +345,19 @@ A navigational link.
       default: 'false',
       description: 'Used to identify the link as the currently active page.',
     },
+    {
+      name: 'onSelect',
+      type: '(event: Event) => void',
+      typeSimple: 'function',
+      description: (
+        <span>
+          Event handler called when the user selects a link (via mouse or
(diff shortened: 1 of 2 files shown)


benoitgrelard (commented):
What's missing here?

andy-hook:
> I can see it's still draft. What's missing here?

Nothing, I was holding off seeking review until the original change was agreed / approved

data/primitives/components/navigation-menu/0.1.3.mdx
```diff
@@ -523,47 +536,41 @@ sure to assign and set a `defaultValue`.
 </NavigationMenu.Root>
 ```
 
-### With router links
+### With client side routing
 
-If you need to use a `Link` provided by your routing package then we recommend composing with `NavigationMenu.Link` via a custom component.
-This will ensure accessibility and consistent keyboard control is maintained. Here's an example using Stitches and React Router:
+If you need to use the `Link` component provided by your routing package then we recommend composing with `NavigationMenu.Link` via a custom component.
+This will ensure accessibility and consistent keyboard control is maintained. Here's an example using Stitches and Next.js:
```
andy-hook, line 542: Anecdotally next.js seems to be more common in requests so I updated the example.

benoitgrelard (approved):

PR #396, opened 2022-05-20
Catchup on demo updates

benoitgrelard:
The only real change was the `data-highlighted` which actually isn't even a breaking change yet, because `:focus` would still work too.


components/demos/Select.jsx
```diff
@@ -115,9 +120,9 @@ const Box = styled('div', {});
 
 export const SelectDemo = () => (
   <Box>
-    <Select defaultValue="blueberry">
+    <Select>
       <SelectTrigger aria-label="Food">
-        <SelectValue />
+        <SelectValue placeholder="Select a fruit…" />
```
benoitgrelard, line 125: I took the opportunity to add placeholder showcase to the demo

data/primitives/components/select/0.1.2.mdx
```diff
@@ -484,6 +484,29 @@ export default () => (
 );
 ```
 
+### With a placeholder
```
benoitgrelard, line 487: as well as an example for it

package.json
```diff
@@ -12,38 +12,37 @@
...
-    "@radix-ui/react-slot": "0.1.2",
-    "@radix-ui/react-switch": "0.1.5",
-    "@radix-ui/react-tabs": "0.1.5",
-    "@radix-ui/react-toast": "^0.1.1",
-    "@radix-ui/react-toggle": "0.1.4",
-    "@radix-ui/react-toggle-group": "0.1.5",
-    "@radix-ui/react-toolbar": "0.1.5",
-    "@radix-ui/react-tooltip": "0.1.7",
-    "@radix-ui/react-visually-hidden": "0.1.4",
+    "@radix-ui/react-accessible-icon": "next",
```
benoitgrelard, line 15: They all needed an update, so all flipped to next

package.json
```diff
@@ -12,38 +12,37 @@
...
-    "@radix-ui/react-collapsible": "0.1.6",
-    "@radix-ui/react-context-menu": "0.1.6",
-    "@radix-ui/react-dialog": "0.1.7",
-    "@radix-ui/react-dropdown-menu": "0.1.6",
-    "@radix-ui/react-hover-card": "0.1.5",
-    "@radix-ui/react-icons": "1.1.0",
-    "@radix-ui/react-label": "0.1.5",
-    "@radix-ui/react-navigation-menu": "^0.1.2",
-    "@radix-ui/react-popover": "0.1.6",
-    "@radix-ui/react-primitive": "0.1.4",
```
benoitgrelard, line 31: I removed this one `react-primitive` as it doesn't seem to be imported anywhere, I'm not sure why it was ever added.

benoitgrelard: I dug, and it looks like it was introduced here: https://github.com/radix-ui/website/commit/3354a5665c36abccbe0b3251ac62cc732a4f98e1 because of `extendPrimitive` but then never removed when we ditched it.

components/demos/Select.jsx
```diff
@@ -20,6 +20,11 @@ const StyledTrigger = styled(SelectPrimitive.SelectTrigger, {
   boxShadow: `0 2px 10px ${blackA.blackA7}`,
   '&:hover': { backgroundColor: mauve.mauve3 },
   '&:focus': { boxShadow: `0 0 0 2px black` },
+  '&[data-placeholder]': { color: violet.violet8 },
```
andy-hook, line 23: I'm finding the contrast here challenging:

image

should we align with the scale recommendation of `11` for low contrast text?

data/primitives/components/select/0.1.2.mdx
```diff
@@ -484,6 +484,29 @@ export default () => (
...
+
+You can use the `placeholder` prop on `Value` for when the select has no value. There's also a `data-placeholder` attribute on `Trigger` to help with styling.
+
+```jsx line=4-6,13
```
andy-hook, line 491:
```suggestion
```jsx line=4-6,11
```
Would that be the intention? right now it looks like:
image

andy-hook (approved):
Nice one

components/demos/Select.jsx
benoitgrelard: haha I know, the issue is 11 is what we have already, so how to differentiate the placeholder from a value then?

data/primitives/components/select/0.1.2.mdx
benoitgrelard: Yep, good catch!

components/demos/Select.jsx
andy-hook: I see what you mean, would `11` for the placeholder and `12` for the active value be appropriate?

benoitgrelard: I tried that but 12 is very dark and looks muddy and out of place as 11 is used everywhere on most components.

benoitgrelard: How about with 9 on the placeholder?
CleanShot 2022-05-20 at 12 21 10@2x

andy-hook: Looks good 👍

PR #407, opened 2022-06-02
Document `data-*` attributes

bukinoshita:

This pull request:

- [ ] Fixes a bug
- [ ] Other


bukinoshita:
> Have a few suggestions:
> 
> * I think we need to remove the space between the prop and data attribute tables so it doesn't feel disjointed.
> * Our tables need a consistent layout, which is especially noticeable with props table and data attribute tables close to each other. I used `width: 40%` on the first column and `width: 30%` on the second and third columns in the examples below.
> 
> Before/after Screenshot 2022-06-07 at 19 24 11 Screenshot 2022-06-07 at 19 25 08

Nice, this looks so much better!

bukinoshita:
@benoitgrelard @vladmoroz Just updated!
screenshot 2022-06-07 at 16 23 43@2x

benoitgrelard:
This is awesome @bukinoshita, I noticed there are a few missing that are burried a bit inside the abstractions, hence why you might have missed them:
- all components using `Popper` also get `data-side`, `data-align` on `Content`
- all components using `Menu` gets the underlying `data-state` too on most parts

I can help go through together if you want.

Also I think your branch should target `release` instead as the base as there has been lots of changes in the docs since (probably new created versions, new parts on menus, etc)

Again, maybe easier if we go through together.

bukinoshita:
> This is awesome @bukinoshita, I noticed there are a few missing that are burried a bit inside the abstractions, hence why you might have missed them:
> 
> * all components using `Popper` also get `data-side`, `data-align` on `Content`
> * all components using `Menu` gets the underlying `data-state` too on most parts
> 
> I can help go through together if you want.
> 
> Also I think your branch should target `release` instead as the base as there has been lots of changes in the docs since (probably new created versions, new parts on menus, etc)
> 
> Again, maybe easier if we go through together.

Yeah, let's go together in the remaining parts.

vladmoroz (approved):
Approved pending Benoît's comments

data/primitives/overview/releases.mdx
```diff
@@ -337,36 +337,36 @@ This release focuses on React 18 support and introduces a number breaking change
 
 <PackageRelease name="Alert Dialog" version="0.0.20" major />
 
-- [**Breaking**] Remove `AlertDialog.Content` `onPointerDownOutside` prop <PRLink id={700} />
+- [**Breaking**] Remove `AlertDialog.Content` `onPointerDownOutside` prop <PRLink id={700} />
```
benoitgrelard, line 340: What happened with this file? Looks like you have a different config maybe for spaces or something? 🤔

data/primitives/components/context-menu/0.1.7.mdx
```diff
@@ -175,6 +175,19 @@ The area that opens the context menu. Wrap it around the target you want the con
   ]}
 />
 
+<DataAttributesTable
+  data={[
+    {
+      attribute: '[data-state]',
+      values: ['open', 'closed'],
+    },
+    {
```
benoitgrelard, line 184: Sorry I should have been clearer. It has `data-state` but not `data-disabled` because it's not a button.

data/primitives/components/context-menu/0.1.7.mdx
```diff
@@ -771,6 +813,23 @@ An item that opens a submenu. Must be rendered inside `ContextMenu.Sub`.
   ]}
 />
 
+<DataAttributesTable
+  data={[
+    {
+      attribute: '[data-state]',
+      values: ['checked', 'unchecked'],
```
benoitgrelard, line 820: I think that would be `open` or `closed` here right?

data/primitives/components/context-menu/0.1.7.mdx
bukinoshita: So we should remove from the `SubTrigger` as well, right?

benoitgrelard: No, subtrigger is always like an item. It's only `ContextMenu.Trigger` that renders a span.

benoitgrelard (approved):

PR #410, opened 2022-06-07
[Select] fix closing tag in docs example

nirtamir2:
- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [ ] Test the change in your own code.
- [ ] Include the URL where we can test the change in the body of your PR.

This pull request:

- [ ] Fixes a bug
- [ ] Adds additional features/functionality
- [x] Updates documentation or example code
- [ ] Other

Fix Select.ItemIndicator closing element in documentation

PR #412, opened 2022-06-08
Sync release with base Toast and Select updates

andy-hook:
Align release with https://github.com/radix-ui/website/pull/410 and https://github.com/radix-ui/website/pull/411


benoitgrelard (approved):
Nice catch!

PR #425, opened 2022-06-23
Add algolia DocSearch

benoitgrelard:
Closes #231

PR #455, opened 2022-08-05
Add algolia custom search

### LINKED ISSUE #231: Add search functionality in documentation
### Examples from other doc sites

https://chakra-ui.com/

benoitgrelard:
This is a continuation of the work started in #425.
We weren't happy with the default built-in experience of DocSearch, so we built our own experience using lower level libraries from Algolia as well as doing a manual scraping of the docs.

Closes #231


components/PrimitivesDocsPage.tsx
```diff
@@ -23,6 +23,17 @@ export function PrimitivesDocsPage({ children }: { children: React.ReactNode })
...
+    const mediaQueryList = window.matchMedia('(min-width: 900px)');
+
+    const handleChange = () => {
+      if (mediaQueryList.matches) setIsMobileMenuOpen(false);
+    };
+
+    mediaQueryList.addEventListener('change', handleChange);
+    return () => mediaQueryList.removeEventListener('change', handleChange);
+  }, []);
+
```
benoitgrelard, line 36: We'll need the same in `ColorsDocsPage` too

components/PrimitivesDocsSearch.tsx
```diff
@@ -222,7 +231,6 @@ export function PrimitivesDocsSearch(props: PrimitivesDocsSearchProps) {
           <TextField
             ref={inputRef}
             {...autocomplete.getInputProps({ inputElement: inputRef.current })}
-            type="search"
```
benoitgrelard: Was that intentional? If there's a reason for it, we should remove the unused CSS too then.

components/PrimitivesDocsSearch.tsx
```diff
@@ -283,54 +291,52 @@ export function PrimitivesDocsSearch(props: PrimitivesDocsSearchProps) {
...
-          )}
-        </DismissableLayer>
+        <RemoveScroll as={Slot} allowPinchZoom enabled={!isMobile}>
+          <DismissableLayer
+            asChild
+            disableOutsidePointerEvents={!isMobile}
+            onPointerDownOutside={(event) => {
+              const target = event.target as HTMLElement;
+              const isPointerDownInForm = formRef.current.contains(target);
+              if (!isMobile && !isPointerDownInForm) autocomplete.setIsOpen(false);
```
benoitgrelard, line 301: 👍

package.json
```diff
@@ -70,7 +71,7 @@
     "prismjs": "^1.23.0",
     "react": "17.0.2",
     "react-dom": "17.0.2",
-    "react-remove-scroll": "^2.4.1",
+    "react-remove-scroll": "^2.5.5",
```
benoitgrelard, line 74: ♥️

components/PrimitivesDocsSearch.tsx
andy-hook: I noticed it's also provided by the prop getter, unless we prefer to redeclare for explicitness?

benoitgrelard: Ah cool, I didn't know that, that's fine if we rely on the prop getter then.

andy-hook (approved):

PR #501, opened 2022-10-24
Add vanilla CSS support in code examples

### LINKED ISSUE #500: Update the docs code blocks to provide both Stitches and CSS examples.
### Why?

To make adoption as easy as possible, we must maximise copy/pasteability. While Stitches is a popular lib, the vast majority of devs still use either plain CSS, Tailwind, or other CSS-in-JS libs. Furthermore, many devs have either never used Stitches, or never even heard of it.

So, when new devs land on the docs site, they are commonly confused by the Stitches code examples. They're often not sure whether these Stitches code snippets are part of Radix, or separate. They don't know that Radix is compatible with any styling solution. So it causes confusion right from the beginning.

By changing the docs code examples to use plain CSS, and provide Stitches an option, we're catering to many more people.

### Solution

Add a language toggle to hero code examples so users can switch between different styling libs like Stitches, CSS, Tailwind etc.

### Language Picker UI

The UI should be similar to the UI for our new Docs. So, a Select menu on the right for choosing the language/lib, and tabs on the left for different files.

Note: Some languages will require just a single file, like Stitches, because both styles and markup are collocated. Other languages will require two files, like CSS, where styles and markup are separated. So the language selector menu should control the tabs UI.

Screenshot 2022-10-18 at 19 33 17

### Notes

- We need to figure out a solution for storing the multiple languages.
- Rewriting the examples in plain CSS should be straightforward. The only variables/tokens used in the code examples are colors. All other properties which accept custom values have their values hard-coded. The colors use imported Radix Colors. So it should be a case of just using the CSS import for Radix Colors instead, and replacing all of the styles with CSS instead of Stitches.
- Language switcher controls code block / tabs UI. Some languages need tabs, others don't.
- It's worth noting that there will be a long-term maintenance burden with this work. When we update a hero code example, or introduce a new component, any changes/updates will need to be made across all languages we support. However, I don't anticipate changes like these often. Just something to be aware of.

### Tailwind

Introducing Tailwind will be a lot trickier.

Tailwind pain points:

- Currently, our examples use a 5px space scale, to accommodate our 15px icons. Tailwind does not support such a space scale. So we will either need to redesign all elements on a 4px space scale, then resize all Radix icons from 15px to 16px, or we'll need to customise the Tailwind config file to use a 5px space scale. If we customise the TW space scale, we lose a lot of copy/pasteability, because close to zero TW users will be using our 5px scale.
- The same goes for TW colors. They use a 10 step scale, Radix Colors uses a 12 step scale. So we will either need to redesign all of our demos to use 10 steps, or extend the TW color scales with two additional steps...again breaking copy/pasteability.

### Approach / Timeline and Tailwind

With the above in mind, I propose we go ahead with the Stitches, CSS, and UI work now, and release it. Then when we're finished, we can take a deeper look at Tailwind.

### Who does this impact? Who is this for?

All docs users.

benoitgrelard:
This PR adds support of vanilla CSS in the docs/code examples (on top of the existing support for Stitches).

Closes #500 

---

### Checklist

- [x] Do a spike on language switching on one chosen component (with both Stitches and vanilla CSS demos/examples)
- [x] Implement styling language switching UI/API #502
- [x] Finish styling tabs/selector UI #531
- [x] Polish collapse/expand UI #532 
- [x] Add `ScrollArea` to codeblock #537
- [x] Re-write all hero demos with vanilla CSS
  - [x] Accordion #503 
  - [x] Alert Dialog #507
  - [x] Aspect Ratio #508
  - [x] Avatar #511 
  - [x] Checkbox #516
  - [x] Collapsible #518 
  - [x] Context Menu #520 
  - [x] Dialog #524 
  - [x] Dropdown Menu #526  
  - [x] Hover Card #527
  - [x] Label #529
  - [x] Navigation Menu #533
  - [x] Popover #530
  - [x] Progress #525
  - [x] Radio Group #522
  - [x] Scroll Area #521
  - [x] Select #519
  - [x] Separator #517
  - [x] Slider #515
  - [x] Switch #514
  - [x] Tabs #510
  - [x] Toast #509
  - [x] Toggle #506
  - [x] Toggle Group #506
  - [x] Toolbar #505
  - [x] Tooltip #504 
- [x] Re-write Getting Started guide with vanilla CSS in mind #535
- [x] Re-write all inline demos (which include CSS changes) with vanilla CSS
- [ ] Re-run Algolia scrape after merge to `main`

---

### End pass

- [x] Check for stitches references throughout (everything should be CSS by default on any page)
- [x] Check there are no inline demos / code blocks with stitches
- [x] Check for `// index.jsx` and `/* styles.css */` comments
- [x] Check for import of `./styles.css`
- [x] Check for consistency around abstracted repetitions vs. straight code
- [x] Check for consistency on imports (general order, color order in css files)


data/primitives/components/context-menu/2.0.1.mdx
```diff
@@ -1221,7 +1221,6 @@ You can add extra decorative elements in the `Item` parts, such as images.
 
 ```jsx line=10,14
 import * as ContextMenu from '@radix-ui/react-context-menu';
-import { CropIcon, TransformIcon } from '@radix-ui/react-icons';
```
benoitgrelard: Should the line have changed too then?

andy-hook: ffs

andy-hook: but thanks ha

andy-hook (approved):

PR #504, opened 2022-10-26
Update Tooltip to use vanilla CSS

benoitgrelard:


components/demos/Tooltip/css/index.jsx
```diff
@@ -0,0 +1,37 @@
...
+import './styles.css';
+
+const TooltipDemo = () => {
+  return (
+    <Tooltip.Provider>
+      <Tooltip.Root>
+        <Tooltip.Trigger asChild>
+          <button className="IconButton">
+            <PlusIcon />
+          </button>
```
benoitgrelard, line 14: For components that don't really matter in the equation, I think this is enough right? Use the native element and stick a class on it, we're not trying to teach how to build a DS.

Thoughts?

andy-hook: This is absolutely fine imo.

components/demos/Tooltip/css/styles.css
```diff
@@ -0,0 +1,101 @@
...
+  padding: 10px 15px;
+  font-size: 15px;
+  line-height: 1;
+  color: var(--violet11);
+  background-color: white;
+  box-shadow: hsl(206 22% 7% / 35%) 0px 10px 38px -10px, hsl(206 22% 7% / 20%) 0px 10px 20px -15px;
+  user-select: none;
+}
+
+@media (prefers-reduced-motion: no-preference) {
```
andy-hook, line 15: Should this be guarding all of our animations? pretty this wasn't applied in our original Accordion demo.

data/primitives/components/tooltip/0.1.1.mdx
```diff
@@ -12,10 +12,6 @@ aria: https://www.w3.org/WAI/ARIA/apg/patterns/tooltip
   receives keyboard focus or the mouse hovers over it.
 </Description>
 
-```jsx hero template=Tooltip
```
andy-hook, line 15: 👍

data/primitives/components/tooltip/1.0.2.mdx
```diff
@@ -512,61 +510,89 @@ export default () => (
...
-    <TooltipContent>…</TooltipContent>
+    <Tooltip.Content __className__="TooltipContent">…</Tooltip.Content>
   </Tooltip.Root>
 );
 ```
 
+```css
+/* styles.css */
+.TooltipContent {
+  transform-origin: var(--radix-tooltip-content-transform-origin);
```
andy-hook, line 528: Should we highlight this line as well as it's pertinent to the overall example?

data/primitives/components/tooltip/1.0.2.mdx
```diff
@@ -512,61 +510,89 @@ export default () => (
...
+.TooltipContent {
+  animation-duration: 0.6s;
+  animation-timing-function: cubic-bezier(0.16, 1, 0.3, 1);
+}
+.TooltipContent[data-side='top'] {
+  animation-name: slideUp;
+}
+.TooltipContent[data-side='bottom'] {
+  animation-name: slideDown;
+}
```
andy-hook, line 571: Thoughts on also highlighting these?

components/demos/Tooltip/css/styles.css
benoitgrelard: To be honest, I'm all for simplifying so i'd be happy to take it out.

data/primitives/components/tooltip/1.0.2.mdx
benoitgrelard: Yeah I wasn't sure about those, given before we were highlighting all the new code necessary for the CSS, but now it's basically the entire CSS file.

Thoughts?

components/demos/Tooltip/css/styles.css
andy-hook: I was more curious if I should be adding it there haha. tbf it does promote a best practice, do you think it gets in the way?

data/primitives/components/tooltip/1.0.2.mdx
andy-hook: I was more thinking about highlighting the parts of this code that are specifically described in the description. In this case:

> We expose a CSS custom property --radix-tooltip-content-transform-origin. Use it to animate the content from its computed origin based on side, sideOffset, align, alignOffset and any collisions.

So it might be helpful to call out where we are using it in the css file.

components/demos/Tooltip/css/styles.css
benoitgrelard: Yeah it's rather that if we do it, we should be consistent. Either never or always. It only really adds 2 lines.

andy-hook: Let's keep it, i'll add it to Accordion as well.

benoitgrelard: Yeah ok, let's say we put those only for top-level demo code.

PR #507, opened 2022-10-27
Add vanilla CSS support to AlertDialog

andy-hook:


components/demos/AlertDialog/css/styles.css
```diff
@@ -0,0 +1,115 @@
...
+.AlertDialogButton.mauve {
+  background-color: var(--mauve4);
+  color: var(--mauve11);
+}
+.AlertDialogButton.mauve:hover {
+  background-color: var(--mauve5);
+}
+.AlertDialogButton.mauve:focus {
+  box-shadow: 0 0 0 2px var(--mauve7);
+}
```
benoitgrelard, line 95: I think this one should just be called `.Button` as it shouldn't be considered an inherent part of the AlertDialog, same as the stitches one using a `Button` component.

benoitgrelard (approved):

PR #508, opened 2022-10-27
Add vanilla CSS support to AspectRatio

andy-hook:


components/demos/AspectRatio/stitches/index.jsx
```diff
@@ -26,10 +15,17 @@ const AspectRatioDemo = () => (
     <AspectRatio.Root ratio={16 / 9}>
       <Img
         src="https://images.unsplash.com/photo-1535025183041-0991a977e25b?w=300&dpr=2&q=80"
-        alt="Landscape photo by Tobias Tullius"
```
andy-hook, line 29: CSB a11y lint rule was complaining about redundant usage of "photo" 🤷‍♂️

benoitgrelard: How about "Landscape photograph…"?

benoitgrelard (approved):

benoitgrelard (approved):

components/demos/AspectRatio/stitches/index.jsx
andy-hook: Kinda silly arbitrary rule but that does pass haha and is better.

PR #509, opened 2022-10-27
Add vanilla CSS support to Toast

benoitgrelard:


components/demos/Toast/css/styles.css
```diff
@@ -0,0 +1,142 @@
...
+.ToastDescription {
+  grid-area: description;
+  margin: 0;
+  color: var(--slate11);
+  font-size: 13px;
+  line-height: 1.3;
+}
+
+.ToastAction {
+  grid-area: action;
```
andy-hook, line 98: It seems this line gets transpiled out to `grid-area: action / action / action / action;` with stitches. At the moment this button isn't aligned centrally on the CSS demo because of this difference.

components/demos/Toast/stitches/index.jsx
```diff
@@ -118,6 +145,12 @@ const Button = styled('button', {
         lineHeight: '25px',
         height: 25,
       },
+      large: {
```
andy-hook, line 148: 👍

data/primitives/components/toast/1.1.1.mdx
```diff
@@ -444,40 +442,43 @@ export default () => {
...
-
-```jsx line=4-7,9-20,24
-import { styled, keyframes } from '@stitches/react';
-import * as ToastPrimitive from '@radix-ui/react-toast';
-
-const slideRight = css.keyframes({
-  from: { transform: 'translateX(var(--radix-toast-swipe-end-x))' },
-  to: { transform: 'translateX(100%)' },
-});
+Combine `--radix-toast-swipe-move-[x|y]` and `--radix-toast-swipe-end-[x|y]` CSS variables with `data-swipe="[start|move|cancel|end]"` attributes to animate a swipe to close gesture. Here's an example using vanilla CSS:
```
andy-hook, line 445:
```suggestion
Combine `--radix-toast-swipe-move-[x|y]` and `--radix-toast-swipe-end-[x|y]` CSS variables with `data-swipe="[start|move|cancel|end]"` attributes to animate a swipe to close gesture. Here's an example:
```

I haven't kept track of other instances of this but I wondered if we need to explicitly mention "vanilla css" or if it's just implicit?

data/primitives/components/toast/1.1.1.mdx
```diff
@@ -444,40 +442,43 @@ export default () => {
...
-    '&[data-swipe="cancel"]': {
-      transform: 'translateX(0)',
-      transition: 'transform 200ms ease-out',
-    },
-    '&[data-swipe="end"]': {
-      animation: `${slideRight} 100ms ease-out`,
-    },
-  },
-});
+```jsx line=4,5
```
andy-hook, line 447:
```suggestion
```jsx line=4,5
// index.jsx
```

data/primitives/components/toast/1.1.1.mdx
benoitgrelard: Yeah same, haven't kept track but you're right, let's not even mention.

components/demos/Toast/css/styles.css
benoitgrelard: I'm not following what you mean here, this is what I see in devtools:

CleanShot 2022-10-31 at 08 50 53@2x

andy-hook: CleanShot 2022-10-31 at 09 04 04

The undo button is offcenter in the css version. I think it's a specificity issue between `unset` and this property.

andy-hook (approved):

PR #519, opened 2022-11-01
Add vanilla CSS support to Select

benoitgrelard:


components/demos/Select/css/index.jsx
```diff
@@ -0,0 +1,84 @@
+import React from 'react';
+import * as Select from '@radix-ui/react-select';
+import classnames from 'classnames';
+import { CheckIcon, ChevronDownIcon, ChevronUpIcon } from '@radix-ui/react-icons';
+import './styles.css';
+
+export const SelectDemo = () => (
```
andy-hook, line 7:
```suggestion
const SelectDemo = () => (
```

components/demos/Select/css/index.jsx
```diff
@@ -0,0 +1,84 @@
...
+);
+
+const SelectContent = React.forwardRef(({ children, className, ...props }, forwardedRef) => {
+  return (
+    <Select.Portal>
+      <Select.Content
+        {...props}
+        ref={forwardedRef}
+        className={classnames('SelectContent', className)}
+      >
```
andy-hook, line 66:
```suggestion
      <Select.Content
        className={classnames('SelectContent', className)}
        {...props}
        ref={forwardedRef}
      >
```

I think we decided to always place `className` first

components/demos/Select/css/index.jsx
```diff
@@ -0,0 +1,84 @@
...
+        {children}
+      </Select.Content>
+    </Select.Portal>
+  );
+});
+
+const SelectItem = React.forwardRef(({ children, className, ...props }, forwardedRef) => {
+  return (
+    <Select.Item {...props} ref={forwardedRef} className={classnames('SelectItem', className)}>
```
andy-hook, line 75:
```suggestion
    <Select.Item className={classnames('SelectItem', className)} {...props} ref={forwardedRef}>
```

components/demos/Select/css/index.jsx
```diff
@@ -0,0 +1,84 @@
...
+      <Select.ItemText>{children}</Select.ItemText>
+      <Select.ItemIndicator>
```
andy-hook, line 77:
```suggestion
      <Select.ItemIndicator className="SelectItemIndicator">
```

I haven't tested if this is what's missing but the indicator is currently on the wrong side:

CleanShot 2022-11-01 at 10 39 58@2x

data/primitives/components/select/1.1.1.mdx
```diff
@@ -542,21 +540,19 @@ Used to visually separate items in the select.
...
-});
-
 export default () => (
   <Select.Root>
     <Select.Trigger>…</Select.Trigger>
     <Select.Portal>
       <Select.Content>
         <Select.Viewport>
-          <StyledItem __disabled__>…</StyledItem>
+          <StyledItem __className__="SelectItem" __disabled__>
```
andy-hook, line 553: These should be `Select.Item`

data/primitives/components/select/1.1.1.mdx
```diff
@@ -566,21 +562,24 @@ export default () => (
...
 import * as Select from '@radix-ui/react-select';
 
-const StyledTrigger = styled(Select.Trigger, {
-  '&[data-placeholder]': { color: 'gainsboro' },
-});
-
 export default () => (
   <Select.Root>
-    <SelectTrigger>
+    <SelectTrigger __className__="SelectTrigger">
```
andy-hook, line 582: `Select.Trigger`

data/primitives/components/select/1.1.1.mdx
```diff
@@ -656,7 +656,7 @@ export default () => (
         <Select.Viewport>
           <Select.Item>
             <Select.ItemText>
-              <Image src="…" />
+              <img src="…" />
```
andy-hook, line 659: I just updated a [similar example](https://github.com/radix-ui/website/pull/520/files#diff-8ebb301faef9380a4588b049166c4364654661942ff0c9eb0fad87aee47f8d1aR1232) and swapped it out to use icons as that seemed a more common use case. Let me know if you'd prefer it to match this or vice versa though.

components/demos/Select/stitches/index.jsx
```diff
@@ -0,0 +1,175 @@
...
+      <SelectViewport>
+        <Select.Group>
+          <SelectLabel>Fruits</SelectLabel>
+          <SelectItem value="apple">Apple</SelectItem>
+          <SelectItem value="banana">Banana</SelectItem>
+          <SelectItem value="blueberry">Blueberry</SelectItem>
+          <SelectItem value="grapes">Grapes</SelectItem>
+          <SelectItem value="pineapple">Pineapple</SelectItem>
+        </Select.Group>
+
```
andy-hook, line 28: Whitespace helps a lot with these longer examples, perhaps I should do that for ContextMenu.

components/demos/Select/css/index.jsx
```diff
@@ -0,0 +1,84 @@
...
+        </Select.Group>
+      </Select.Viewport>
+      <Select.ScrollDownButton className="SelectScrollButton">
+        <ChevronDownIcon />
+      </Select.ScrollDownButton>
+    </SelectContent>
+  </Select.Root>
+);
+
+const SelectContent = React.forwardRef(({ children, className, ...props }, forwardedRef) => {
```
andy-hook, line 59: I went the other way with ContextMenu and removed the abstractions, do you think it'd be better to keep it as you have here? I could go either way tbh.

data/primitives/components/select/1.1.1.mdx
benoitgrelard: I think I prefer img for the simplicity

components/demos/Select/stitches/index.jsx
benoitgrelard: With the groups it helps yeah.

components/demos/Select/css/index.jsx
andy-hook: > I think maybe we don't abstract what's not repeated (so for example, I should remove the portal/content abstraction) but for repeated items we do?

This sounds good.

benoitgrelard:
All done 👍

andy-hook (approved):
Looks great 💯

PR #520, opened 2022-11-01
Add vanilla css support to ContextMenu

andy-hook:


data/primitives/components/context-menu/2.0.1.mdx
```diff
@@ -1215,33 +1217,24 @@ export default () => {
...
 
 export default () => (
   <ContextMenu.Root>
     <ContextMenu.Trigger>…</ContextMenu.Trigger>
     <ContextMenu.Portal>
       <ContextMenu.Content>
         <ContextMenu.Item>
-          <Image src="…" />
-          Adolfo Hess
+          <TransformIcon />
```
benoitgrelard, line 1232: I just put `` in my PR I think that's enough right?

andy-hook: ha, yes I raised a similar comment on yours. Let's go with `img` here as you have.

benoitgrelard (approved):

PR #523, opened 2022-11-02
Tweaking styles for new code block

colmtuite:

This pull request:

- [ ] Fixes a bug
- [ ] Other

### DIFF
--- a/components/HeroCodeBlock.tsx
+++ b/components/HeroCodeBlock.tsx
@@ -89,7 +89,13 @@ export const HeroCodeBlock = ({ children }: { children?: React.ReactNode }) => {
         <Collapsible.Content asChild forceMount>
           <Box css={{ position: 'relative' }}>
             <Tabs.Root value={currentTabValue} onValueChange={setCurrentTabValue}>
-              <Flex css={{ backgroundColor: '$violet3', padding: '$2' }}>
+              <Flex
+                css={{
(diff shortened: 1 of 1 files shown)


benoitgrelard:
Merging this in.

andy-hook:
Would `ScrollArea` be contender for handling the overflow? the current gradient overlay on the scrollbar interferes with the interaction and doesn't look great.

benoitgrelard:
> Would ScrollArea be a contender for handling the overflow?

Yep, I will look at this. 

> the current gradient overlay on the scrollbar interferes with the interaction and doesn't look great.

I agree, I raised to Colm but he ran out of time. I'll ask Vlad if he has some time.

andy-hook:
> I'll ask Vlad if he has some time.

I think the gradient is good as an overflow affordance (it just looks a bit odd on top of the native bar) if we can slot a custom bar over the top that'd probably be enough. Happy to help if needed.

benoitgrelard:
What do you call "the native bar" here?

andy-hook:
CleanShot 2022-11-04 at 10 28 14

benoitgrelard:
ah yeah ok, I've added a check in the list to add scroll area.
My main issue with the current collapse/expand UI is that it's not very accessible, can't really see the button, it's muddy on top of the text behind, etc.

PR #527, opened 2022-11-03
Add vanilla CSS support to HoverCard 

andy-hook:


components/demos/HoverCard/stitches/index.jsx
```diff
@@ -33,38 +77,17 @@ const StyledContent = styled(HoverCardPrimitive.Content, {
...
-    '&[data-state="open"]': {
-      '&[data-side="top"]': { animationName: slideDownAndFade },
-      '&[data-side="right"]': { animationName: slideLeftAndFade },
-      '&[data-side="bottom"]': { animationName: slideUpAndFade },
-      '&[data-side="left"]': { animationName: slideRightAndFade },
-    },
+    '&[data-side="top"]': { animationName: slideDownAndFade },
+    '&[data-side="right"]': { animationName: slideLeftAndFade },
+    '&[data-side="bottom"]': { animationName: slideUpAndFade },
+    '&[data-side="left"]': { animationName: slideRightAndFade },
```
benoitgrelard, line 83: Why this change?
The recommendation should be to qualify it right?

benoitgrelard (approved):

components/demos/HoverCard/stitches/index.jsx
andy-hook: Yes you're right, I must have tripped. Presumably the examples should match?

e.g. https://www.radix-ui.com/docs/primitives/components/dropdown-menu#collision-aware-animations

benoitgrelard: Yeah I'll add an item to the last pass checklist to go through for consistency.

andy-hook: > not sure if you saw but I added a last pass checklist

I didn't but this is good stuff, was going to suggest similar. I'll revert this particular change and we'll address others there.

benoitgrelard (approved):

PR #531, opened 2022-11-04
Tweak tabs UI styles

benoitgrelard:
I've added better focus accessibility by bringing it inline with what is done in the WorkOS docs, I think it works very well.
Also slightly updated the Select we use internally to bring it close with the DS ghost buttons which brings things more inline.

### DIFF
--- a/components/HeroCodeBlock.tsx
+++ b/components/HeroCodeBlock.tsx
@@ -129,22 +129,37 @@ export const HeroCodeBlock = ({ children }: { children?: React.ReactNode }) => {
                           '&[data-state="active"]': {
                             fontWeight: 500,
                             letterSpacing: '-.025em',
-                            '&:focus': {
-                              '&::before': {
-                                boxSizing: 'border-box',
-                                content: '""',
-                                height: '2px',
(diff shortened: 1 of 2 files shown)


components/HeroCodeBlock.tsx
```diff
@@ -178,7 +194,7 @@ export const HeroCodeBlock = ({ children }: { children?: React.ReactNode }) => {
                       borderBottomRightRadius: '$3',
                       '&:focus': {
                         outline: 'none',
-                        boxShadow: '0 0 0 2px $colors$slateA8',
+                        boxShadow: '0 0 0 2px $colors$violet7',
```
andy-hook, line 197: Should we the match the existing color that is used for all our other focus rings? not a biggie just felt inconsistent when tabbing across these controls.

andy-hook (approved):

components/HeroCodeBlock.tsx
benoitgrelard: I think it looks better with the violet considering the code block is all in this tone.

PR #542, opened 2022-11-11
Fix SSR rendering in demos

andy-hook:
Our styles were not rendering in SSR.

CleanShot 2022-11-11 at 08 48 16


lib/remark-hero-template.js
```diff
@@ -1,29 +0,0 @@
-const fs = require('fs');
```
benoitgrelard, line 1: nice 😀

benoitgrelard (approved):
Ohhh so that would be why we had that then! 😀

PR #546, opened 2022-11-15
November 15 release

andy-hook:


benoitgrelard (approved):

benoitgrelard:
We need to update the deps also right?

andy-hook:
> We need to update the deps also right

Probably should yeah https://github.com/radix-ui/website/pull/547

Issue #540, opened 2022-11-10
Add tailwind demos


benoitgrelard:
We want to add Tailwind demos alongside CSS and Stitches.

- [x] Accordion @joaom00  #557 
- [x] Alert Dialog @shadcn #603 
- [x] Aspect Ratio @shadcn #579 
- [x] Avatar @shadcn #580
- [x] Checkbox @shadcn #577
- [x] Collapsible @ryanleichty #595 
- [x] Context Menu @ryanleichty #593 
- [x] Dialog @shadcn @shadcn #602 
- [x] Dropdown Menu @ryanleichty #592 
- [x] Hover Card @shadcn #578
- [x] Label @shadcn #581
- [x] Menubar @ryanleichty #594 
- [x] Navigation Menu @shadcn #604
- [x] Popover @joaom00 #576 
- [x] Progress @joaom00 #582 
- [x] Radio Group @joaom00 #583 
- [x] Scroll Area @joaom00 #584 
- [x] Select @ryanleichty #585 
- [x] Separator @joaom00 #586 
- [x] Slider @joaom00 #587 
- [x] Switch @joaom00 #588 
- [x] Tabs @joaom00 #589 
- [x] Toast @ryanleichty #596 
- [x] Toggle @suliskh #590
- [x] Toggle Group @suliskh #591 
- [x] Toolbar @ryanleichty #599
- [x] Tooltip @shadcn #601

romansndlr:
Yes! We use Radix UI for our design system at Venn and we want to help! Since we use Tailwind this can be a great way for us to give back. How do we start?

benoitgrelard:
Hey Roman, that's amazing! What a great way to give back! 🥰

Here are a few things to consider:
- One thing we're unsure how we'll achieve is how to closely match the Tailwind demos to what we currently have
  - spacing/colors might differ from the values we used in the demos (we use a 5px scale)
  - when we use icons, our icons have been designed at 15px (so fit nicely with our 5px scale)
- Perhaps we can find a way to include a customised tailwind config that matches?
- Or perhaps it's a better experience for users to see an out of the box Tailwind config, and it doesn't matter that the colors/size might look a tiny bit different

---

Perhaps a good starting point would be to do a **POC with one of the primitives**. Would you like to pick one and run with it?

I think roughly this is what it would involve:
- Add `tailwind` as a CSS lib in https://github.com/radix-ui/website/blob/main/lib/constants.ts
[...]

PR #557, opened 2023-01-03
[POC][Accordion] Add tailwind CSS support

joaom00:
Some points

- I wasn't able to use `create-react-app` as template on codesandbox to run tailwindcss so I had to use `vite`. If you guys have any idea how to setup with `create-react-app` I appreciate
- I think we can put the space too that is used in demos in `tailwind.config.js`. I don't know if is too much for the user to copy?

This pull request:

- [ ] Fixes a bug
- [ ] Other

### DIFF
--- a/components/HeroCodeBlock.tsx
+++ b/components/HeroCodeBlock.tsx
@@ -307,37 +307,44 @@ svg {
     [`@radix-ui/react-${componentName}`]: 'latest',
   };
 
+  const devDependencies = {};
+
   if (cssLib === 'css') {
     dependencies['classnames'] = 'latest';
+    devDependencies['react-scripts'] = 'latest';
   }
 
   if (cssLib === 'stitches') {
     dependencies['@stitches/react'] = 'latest';
+    devDependencies['react-scripts'] = 'latest';
   }
 
-  const files = {
-    'package.json': {
-      content: {
-        dependencies,
-        devDependencies: {
-          'react-scripts': 'latest',
-        },
-      } as any,
-      isBinary: false,
-    },
-    'App.js': {
-      content: sources['index.jsx'],
-      isBinary: false,
-    },
-    'index.js': {
-      content: `import { createRoot } from 'react-dom/client';
+  if (cssLib === 'tailwind') {
+    dependencies['classnames'] = 'latest';
+    devDependencies['vite'] = 'latest';
+    devDependencies['@vitejs/plugin-react'] = 'latest';
+    devDependencies['tailwindcss'] = 'latest';
+    devDependencies['postcss'] = 'latest';
+    devDependencies['autoprefixer'] = 'latest';
+  }
+
+  const appFileOptions = { content: sources['index.jsx'], isBinary: false };
+  const indexContent = `import { createRoot } from 'react-dom/client';
 import App from './App';
 import './global.css';
 
 const container = document.getElementById('root');
 const root = createRoot(container);
 
-root.render(<App />);`,
+root.render(<App />);`;
+  const indexFileOptions = { content: indexContent, isBinary: false };
+
+  const files = {
+    'package.json': {
+      content: {
+        dependencies,
+        devDependencies,
+      } as any,
       isBinary: false,
     },
     'global.css': {
@@ -347,11 +354,82 @@ root.render(<App />);`,
   };
 
   if (cssLib === 'css') {
+    files['App.js'] = appFileOptions;
+    files['index.js'] = indexFileOptions;
     files['styles.css'] = {
       content: sources['styles.css'],
       isBinary: false,
     };
   }
 
-  return getParameters({ files });
+  if (cssLib === 'stitches') {
+    files['App.js'] = appFileOptions;
+    files['index.js'] = indexFileOptions;
+  }
+
+  if (cssLib === 'tailwind') {
+    files['App.jsx'] = appFileOptions;
+    files['index.jsx'] = indexFileOptions;
+    files['tailwind.config.js'] = {
+      content: sources['tailwind.config.js'],
+      isBinary: false,
+    };
+    files['postcss.config.js'] = {
+      content: `module.exports = {
+  plugins: {
+    tailwindcss: {},
+    autoprefixer: {},
+  }
+}`,
+      isBinary: false,
+    };
+    files['package.json']['content']['scripts'] = {
+      start: 'vite',
+    };
+    files['global.css'] = {
+      content: `@tailwind base;
+@tailwind components;
+@tailwind utilities;
+
+body {
+  font-family: system-ui;
+  width: 100vw;
+  height: 100vh;
+  background-image: linear-gradient(330deg, hsl(272, 53%, 50%) 0%, hsl(226, 68%, 56%) 100%);
+  display: flex;
+  align-items: flex-start;
+  justify-content: center;
+  margin-top: 120px;
+}`,
+      isBinary: false,
+    };
+    files['vite.config.js'] = {
+      content: `import { defineConfig } from 'vite'
+import react from '@vitejs/plugin-react'
+
+// https://vitejs.dev/config/
+export default defineConfig({
+  plugins: [react()],
+})`,
+      isBinary: false,
+    };
+    files['index.html'] = {
+      content: `<!DOCTYPE html>
+    <html lang="en">
+      <head>
+        <meta charset="UTF-8" />
+        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+        <title>Vite App</title>
+      </head>
+      <body>
+        <div id="root"></div>
+        <script type="module" src="/index.jsx"></script>
+      </body>
+    </html>`,
+      isBinary: false,
+    };
+  }
+
+  const template = cssLib === 'tailwind' ? 'node' : 'create-react-app';
+  return getParameters({ files, template });
 };
--- a/components/HeroCodeBlock.tsx
+++ b/components/HeroCodeBlock.tsx
@@ -24,7 +24,6 @@ export const HeroCodeBlock = ({
   const [preferredCssLib, setPreferredCssLib] = useCssLibPreference();
   const usedCssLib = cssLibProp ?? preferredCssLib;
   const [isCodeExpanded, setIsCodeExpanded] = React.useState(false);
-  const module = usedCssLib === 'tailwind' ? 'App.jsx' : 'App.js';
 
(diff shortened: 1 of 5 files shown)


joaom00:
@benoitgrelard @andy-hook sorry for the ping 😅

Can you guys take a look at this so we can take it forward?

benoitgrelard:
> I wasn't able to use create-react-app as template on codesandbox to run tailwindcss so I had to use vite. If you guys have any idea how to setup with create-react-app I appreciate

I found [this](https://tailwindcss.com/docs/guides/create-react-app) with regards to setting up tailwind with create-react-app, although they do recommend other options.

benoitgrelard:
Not sure if it's just me or codesandbox, but it's kinda stuck on this screen:
CleanShot 2023-01-27 at 13 07 30@2x

andy-hook:
> Not sure if it's just me or codesandbox, but it's kinda stuck on this screen:

I've found CSB containers to be very buggy, they rarely start for me and when they do they fail with errors on return visits.

components/HeroCodeBlock.tsx
```diff
@@ -24,6 +24,7 @@ export const HeroCodeBlock = ({
   const [preferredCssLib, setPreferredCssLib] = useCssLibPreference();
   const usedCssLib = cssLibProp ?? preferredCssLib;
   const [isCodeExpanded, setIsCodeExpanded] = React.useState(false);
+  const module = usedCssLib === 'tailwind' ? 'App.jsx' : 'App.js';
```
benoitgrelard, line 27: What is the reason for this?
Does Vite enforce the use of `.jsx`?
Could we unify to always be `App.js` so it simplifies things?

components/HeroCodeBlock.tsx
```diff
@@ -347,11 +355,82 @@ root.render(<App />);`,
...
+    files['App.js'] = appFileOptions;
+    files['index.js'] = indexFileOptions;
     files['styles.css'] = {
       content: sources['styles.css'],
       isBinary: false,
     };
   }
 
-  return getParameters({ files });
+  if (cssLib === 'stitches') {
```
benoitgrelard, line 366: Now that we're growing to 3 solutions, this whole section is starting to be a bit hard to read with all the branching, etc.

I think that would be a good candidate to break apart in multiple functions.

components/demos/Accordion/tailwind/tailwind.config.js
```diff
@@ -0,0 +1,50 @@
...
+        violet3: 'hsl(252, 96.9%, 97.4%)',
+        violet4: 'hsl(252, 91.5%, 95.5%)',
+        violet5: 'hsl(252, 85.1%, 93.0%)',
+        violet6: 'hsl(252, 77.8%, 89.4%)',
+        violet7: 'hsl(252, 71.0%, 83.7%)',
+        violet8: 'hsl(252, 68.6%, 76.3%)',
+        violet9: 'hsl(252, 56.0%, 57.5%)',
+        violet10: 'hsl(251, 48.1%, 53.5%)',
+        violet11: 'hsl(250, 43.0%, 48.0%)',
+        violet12: 'hsl(254, 60.0%, 18.5%)',
```
benoitgrelard, line 31: We should add `@radix-ui/colors` as a dependency like in the other solutions (Stitches or CSS) and import the colors here instead.

benoitgrelard: Especially if we're going to have a copy of the tailwind config per demo, we'll want to minimize the amount of discprancies between.

lib/rehype-hero-code-block.ts
```diff
@@ -14,10 +14,10 @@ const rehypeHeroCodeBlock = () => (tree: UnistTree) => {
         node.children = [];
 
         SUPPORTED_CSS_LIBS.forEach((lib) => {
-          ['index.jsx', 'styles.css'].forEach((file) => {
+          ['index.jsx', 'styles.css', 'tailwind.config.js'].forEach((file) => {
             const filePath = `${process.cwd()}/components/demos/${folder}/${lib}/${file}`;
             if (fileExists(filePath)) {
-              const [, extension] = file.split('.');
+              const [extension] = file.split('.').slice(-1);
```
benoitgrelard, line 20:
```suggestion
              const extension = file.split('.').pop();
```

benoitgrelard (commented):
I can't see the demo in the sandbox so can't comment yet on that, but the rest is looking promising!

benoitgrelard:
> > Not sure if it's just me or codesandbox, but it's kinda stuck on this screen:
> 
> I've found CSB containers to be very buggy, they rarely start for me and when they do they fail with errors on return visits.

What's the alternative?

benoitgrelard:
@joaom00 the demo looks good, the only thing I can see is it seems to be missing some focus outline reset:

CleanShot 2023-01-27 at 13 26 23@2x

andy-hook:
> And after that it was indeed much faster in the new codesandbox editor.
Is it possible to target this from the start?

It seems you can access cloud editor projects by spinning up a specific cloud template. I wonder if we can pass a cloud template name to `getParameters` to achieve the same 🤔

benoitgrelard:
I haven't found anything about that here: https://codesandbox.io/docs/learn/getting-started/your-first-sandbox#define-api

components/HeroCodeBlock.tsx
joaom00: > Could we unify to always be App.js so it simplifies things?

We can, but we need a [different configuration in vite](https://github.com/vitejs/vite/discussions/3448)

andy-hook:
> I haven't found anything about that here: https://codesandbox.io/docs/learn/getting-started/your-first-sandbox#define-api

CleanShot 2023-01-27 at 13 57 19

Looks like we're limited. I've asked in their discord.

components/HeroCodeBlock.tsx
benoitgrelard: Ok, not worth drifting away from the most common way the tool works then.

components/HeroCodeBlock.tsx
```diff
@@ -79,7 +80,13 @@ export const HeroCodeBlock = ({
             method="POST"
             target="_blank"
           >
-            <input type="hidden" name="query" value="module=App.js" />
+            <input type="hidden" name="query" value={`file=/${module}`} />
```
joaom00, line 83: I need to use `file` because `module` doesn't seem to work on cloud sandboxes

components/HeroCodeBlock.tsx
```diff
@@ -79,7 +80,13 @@ export const HeroCodeBlock = ({
...
+            {usedCssLib === 'tailwind' && (
+              <>
+                <input type="hidden" name="environment" value="server" />
+                <input type="hidden" name="hidedevtools" value="1" />
+              </>
+            )}
```
joaom00, line 89: Cloud sandboxes don't seem to work with `create-react-app` template

joaom00:
The tailwind demo is now using cloud sandboxes

benoitgrelard (commented):
Oh so neat you got the cloud sandbox working!
I am wondering if we could / should make it so it's always in this environment for all (even stitches/css), I feel like the new experience is much nicer!

@andy-hook what do you think about that?

andy-hook:
> Oh so neat you got the cloud sandbox working! I am wondering if we could / should make it so it's always in this environment for all (even stitches/css), I feel like the new experience is much nicer!

I agree, the new editor is a huge improvement! Providing it's safe to do so I think we should switch to this for everything.

joaom00:
+1 on switching to the new editor. I should make on this PR or would it be better to do it in another PR?

benoitgrelard:
Let's do it on this PR!

joaom00:
@benoitgrelard Just a note, I would have to use vite in the other solutions too

benoitgrelard:
> @benoitgrelard Just a note, I would have to use vite in the other solutions too

Yeah I think that's ok, as the config is super minimal and out of the way.

components/HeroCodeBlock.tsx
```diff
@@ -307,51 +314,154 @@ svg {
...
-  if (cssLib === 'css') {
-    files['styles.css'] = {
-      content: sources['styles.css'],
-      isBinary: false,
-    };
+  if (cssLib === 'css') makeCssConfig(dependencies, devDependencies, sources, files);
+  if (cssLib === 'stitches') makeStitchesConfig(dependencies, devDependencies, sources, files);
+  if (cssLib === 'tailwind') makeTailwindConfig(dependencies, devDependencies, sources, files);
+
+  const template = cssLib === 'tailwind' ? 'node' : 'create-react-app';
```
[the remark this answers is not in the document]
joaom00: Yes 😅

components/HeroCodeBlock.tsx
```diff
@@ -307,51 +314,154 @@ svg {
...
   };
 
-  if (cssLib === 'css') {
-    files['styles.css'] = {
-      content: sources['styles.css'],
-      isBinary: false,
-    };
+  if (cssLib === 'css') makeCssConfig(dependencies, devDependencies, sources, files);
+  if (cssLib === 'stitches') makeStitchesConfig(dependencies, devDependencies, sources, files);
+  if (cssLib === 'tailwind') makeTailwindConfig(dependencies, devDependencies, sources, files);
```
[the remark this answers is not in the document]
joaom00: Done. I think... 😅

andy-hook (approved):

benoitgrelard (approved):

benoitgrelard:
Thanks for kicking this off @joaom00!

Now that this has been merged 🎉 we can all work on adding the missing TW demos.

I have updated the issue #540 with a list, I suggest the way we do this is:
- when someone has time, pick one component
- put your name against it in the issue so we know not to step on each other's toes
- when PR is raised, link the PR in the issue next to the component

What do you all think?

cc @andy-hook

andy-hook:
> put your name against it in the issue so we know not to step on each other's toes

Were we thinking of editing the issue description to do this? or just posting a comment. Only @joaom00 wouldn't be able to edit the list in the desc afaik.

benoitgrelard:
Ah good point, @joaom00 let me know if you can't edit the description.
And if so no worries, just post a comment on the issue telling us which one you pickup and I'll update the description accordingly.

joaom00:
@benoitgrelard I can't edit the issue description

benoitgrelard:
Ok no problem, are you ok with adding a comment when you pick one up then and I'll update the description accordingly?

joaom00:
Yes sure :+1:

PR #571, opened 2023-02-06
[Slider] Add styles to vertical example

joaom00:
Fix slider styles don't work correctly when orientation is vertical

This pull request:

- [ ] Fixes a bug
- [ ] Other

### DIFF
--- a/data/primitives/components/slider/1.1.0.mdx
+++ b/data/primitives/components/slider/1.1.0.mdx
@@ -307,25 +307,46 @@ export default () => (
     <Slider.Track className="SliderTrack">
       <Slider.Range className="SliderRange" />
     </Slider.Track>
-    <Slider.Thumb />
+    <Slider.Thumb className="SliderThumb" />
   </Slider.Root>
 );
 ```
 
-```css
+```css line=7,18,26
 /* styles.css */
-.SliderRoot[data-orientation='vertical'] {
+.SliderRoot {
+  position: relative;
+  display: flex;
+  align-items: center;
+}
+.SliderRoot[__data-orientation='vertical'__] {
   flex-direction: column;
   width: 20px;
   height: 100px;
 }
 
-.SliderTrack[data-orientation='vertical'] {
+.SliderTrack {
+  position: relative;
+  flex-grow: 1;
+  background-color: grey;
+}
+.SliderTrack[__data-orientation='vertical'__] {
   width: 3px;
 }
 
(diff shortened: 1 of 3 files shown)


benoitgrelard (commented):
We usually only include styles relating to the demo itself. For examples in the tabs demo, we don't include vertical orientation either.

What do you think @andy-hook?

andy-hook:
> What do you think

I agree, the hero demo should only contain the styles it needs. Would could expand the [vertical orientation](https://www.radix-ui.com/docs/primitives/components/slider#vertical-orientation) example with a style block showing this instead?

benoitgrelard:
> We could expand the [vertical orientation](https://www.radix-ui.com/docs/primitives/components/slider#vertical-orientation) example with a style block showing this instead?

Ah yes great idea!

joaom00:
Should I include the style block only in latest version?

andy-hook:
> Should I include the style block only in latest version?

Just the latest should be fine 👍

data/primitives/components/slider/1.1.0.mdx
```diff
@@ -293,19 +293,42 @@ A draggable thumb. You can render multiple thumbs.
...
+  height: 100px;
+}
+
+.SliderTrack[data-orientation='vertical'] {
+  width: 3px;
+}
+
+.SliderRange[data-orientation='vertical'] {
+  width: 100%;
+}
```
andy-hook, line 329: Let's align the track inside the root and add functional styles for the thumb as well, at this point we may as well make this a (simple) functioning demo.

andy-hook:
Thank you for this @joaom00 , just one more small comment then we can get this in.

data/primitives/components/slider/1.1.0.mdx
joaom00: So should I add the same styles from the demo + these one?

andy-hook: Just the styles needed to make it functional so that if we copied the code it would give at least a working example:

https://codesandbox.io/s/simple-vertical-slider-dl45lc?file=/styles.css

And then if we could highlight the `data-orientation` lines to show that vertical is a variation on the base style.

Let me know if that makes sense!

joaom00: Yeah, that makes sense!

andy-hook (approved):
Great stuff 🌟

PR #573, opened 2023-02-06
Fix React import in getting started examples

### LINKED ISSUE #572: Wrong React import in code examples
### Relevant Radix Component(s)

The React import is wrong in the code examples on the [getting started page](https://www.radix-ui.com/docs/primitives/overview/getting-started). It should be `'react'`, not `'React'`.

I could make a PR to fix it on this page but it would probably be best to check the whole project for this mistake.

Also, is the import even necessary? Back in the day we needed to import React for JSX to work, but in modern projects it should't be an issue I think? If I paste the code in a `.jsx` or `.tsx` file, the editor complains about an unused variable.

tomek-ch:
fixes #572

### DIFF
--- a/data/primitives/overview/getting-started.mdx
+++ b/data/primitives/overview/getting-started.mdx
@@ -28,7 +28,7 @@ Import and structure the parts.
 
 ```jsx line=3,6-14
 // index.jsx
-import * as React from 'React';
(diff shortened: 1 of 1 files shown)


andy-hook (approved):
Thank you ⭐

PR #590, opened 2023-02-10
[Tailwind] Add tailwind demo for Toggle

suliskh:

This pull request:

- [ ] Fixes a bug
- [ ] Other

### DIFF
--- a/components/demos/Toggle/tailwind/index.jsx
+++ b/components/demos/Toggle/tailwind/index.jsx
@@ -5,7 +5,7 @@ import { FontItalicIcon } from '@radix-ui/react-icons';
 const ToggleDemo = () => (
   <Toggle.Root
     aria-label="Toggle italic"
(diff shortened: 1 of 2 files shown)


suliskh:
Halo @benoitgrelard @andy-hook, I made this Tailwind example to be as idiomatic as possible. As a tradeoff, it'll have slight differences (e.g. `width`, `height`, `box-shadow`) with the CSS example. 

Please let me know what's your opinion on that. Thank you 👍🏻

benoitgrelard:
> Halo @benoitgrelard @andy-hook, I made this Tailwind example to be as idiomatic as possible. As a tradeoff, it'll have slight differences (e.g. `width`, `height`, `box-shadow`) with the CSS example.
> 
> Please let me know what's your opinion on that. Thank you 👍🏻

@suliskh For the sake of consistency for now, let's make it the same as all the other demos by using arbitrary values that match as we haven't really decided yet if we want to make those more idiomatic by using default TW colors, sizes, etc. This is primarily because our icon system is based on a 15px size and all our sizes are increment of 5px which differs from TW.

But yeah I'd say for now let's make it the same as the other demos, and we can decide later if we want to move all the TW to something more idiomatic (even if they end up with slight differences).

benoitgrelard (approved):
♥️

PR #597, opened 2023-02-12
Add antialiasing to Tailwind config

ryanleichty:
I noticed the fonts were lacking antialiasing in the Tailwind codesandbox, whereas the preview on the Radix website has it. I added this to the Tailwind config. Normally I add the Tailwind `antialiased` class directly to the body tag

This pull request:

- [ ] Fixes a bug
- [ ] Other

### DIFF
--- a/components/HeroCodeBlock.tsx
+++ b/components/HeroCodeBlock.tsx
@@ -352,6 +352,8 @@ body {
   align-items: flex-start;
   justify-content: center;
   margin-top: 120px;
+  -webkit-font-smoothing: antialiased;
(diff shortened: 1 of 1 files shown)


benoitgrelard (commented):
Yeah I noticed that too. It's actually missing from all the demos, not just the TW ones. Would you mind adding to all of the global.css files? (so also stitches and CSS)

ryanleichty:
Good catch! I didn't check those ones. I just updated it, so all three demos should have antialiasing now.

benoitgrelard (approved):

PR #611, opened 2023-03-07
Add `Form` docs

benoitgrelard:
This PR adds docs for the new `Form` primitive.

### DIFF
--- a/data/primitives/components/form/0.0.1.mdx
+++ b/data/primitives/components/form/0.0.1.mdx
@@ -434,7 +434,7 @@ The field is marked as invalid by passing a `serverInvalid` boolean prop to the 
 
 Here's an example with server-side error handling:
 
-```jsx line=12,23,29-34
+```jsx line=5-8,13,23,28-30,49-54
 import * as React from 'react';
 import * as Form from '@radix-ui/react-form';
 
@@ -457,13 +457,13 @@ function Page() {
            * Map errors from your server response into a structure you'd like to work with.
            * In this case resulting in this object: `{ email: false, password: true }`
            */
-          .catch((errors) => setServerErrors(mapServerErrors(errors)));
+          .catch((errors) => __setServerErrors__(mapServerErrors(errors)));
 
         // prevent default form submission
         event.preventDefault();
       }}
       onClearServerErrors={() =>
-        setServerErrors({ email: false, password: false })
+        __setServerErrors__({ email: false, password: false })
       }
     >
       <Form.Field name="email" __serverInvalid__={serverErrors.email}>
--- a/data/primitives/components/form/0.0.1.mdx
+++ b/data/primitives/components/form/0.0.1.mdx
@@ -199,7 +199,7 @@ A control element (by default an `input`) which is automatically wired when nest
 
 A validation message which is automatically wired (functionality and accessibility) to a given control when nested inside a `Field` part. It can be used for built-in and custom client-side validation, as well as server-side validation. When used outside a `Field` you must pass a `name` prop matching a field.
 
-`Form.Message` accepts a `match` prop which is used to determine when the message should show. It matches the native HTML validity state (`ValidityState` on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState)) which result from using attributes such as `required`, `min`, `max`. The message will show if the given `match` is `true` on the control’s validity state.
+`Form.Message` accepts a `match` prop which is used to determine when the message should show. It matches the native HTML validity state (`ValidityState` on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState)) which validates against attributes such as `required`, `min`, `max`. The message will show if the given `match` is `true` on the control’s validity state.
 
 You can also pass a function to `match` to provide custom validation rules.
 
@@ -259,8 +259,7 @@ You can also pass a function to `match` to provide custom validation rules.
       type: 'string',
       description: (
         <>
-          When used outside of a <Code>Field</Code> part, use this prop to
-          target a specific field by name.
+          Used to target a specific field by name when rendering outside of a <Code>Field</Code> part.
         </>
       ),
     },
@@ -269,7 +268,7 @@ You can also pass a function to `match` to provide custom validation rules.
 
 ### ValidityState
 
-Use this render-prop component to access a given field’s validity state in render (see `ValidityState` on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState)). When nested inside a `Field` part, it’s automatic, otherwise you must pass a `name` prop matching a field.
+Use this render-prop component to access a given field’s validity state in render (see `ValidityState` on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState)). A field's validity is available automatically when nested inside a `Field` part, otherwise you must pass a `name` prop to associate it.
 
 <PropsTable
   data={[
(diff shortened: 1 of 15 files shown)


data/primitives/components/form/0.0.1.mdx
```diff
@@ -0,0 +1,520 @@
...
+You can provide a more meaningful message by providing your own `children`. This is also useful for internationalization.
+
+```jsx
+// will yield "Please provide a name"
+<Form.Message match="missingValue">__Please provide a name__</Form.Message>
+```
+
+### Custom validation
+
+On top of all the built-in client-side validation matches described above you can also provide your own custom validation whilst still making use of the platform's validation abilities. It uses the `customError` type present in the constraint validity API.
```
andy-hook, line 361: "constraint validity API" can we link to it?

data/primitives/components/form/0.0.1.mdx
```diff
@@ -0,0 +1,520 @@
...
+
+You can pass your own validation function into the `match` prop on `Form.Message`. Here's a contrived example:
```
andy-hook, line 363:
```suggestion
You can pass your own validation function into the `match` prop on `Form.Message`. Here's an example:
```
nit: imo contrived has a negative connotation without adding much.

data/primitives/components/form/0.0.1.mdx
```diff
@@ -0,0 +1,520 @@
...
+  <Form.Label>Full name</Form.Label>
+  <Form.Control />
+  <Form.Message __match__={(value, formData) => value !== 'John'}>
+    Only John is allowed.
+  </Form.Message>
+</Form.Field>
+```
+
+> `match` will be called with the current value of the control as first argument and the entire `FormData` as second argument.
+> `match` can also be an `async` function (or return a promise) to perform async validation (see type below):
```
andy-hook, line 376: "see type below" – what is this referring to?

data/primitives/components/form/0.0.1.mdx
```diff
@@ -0,0 +1,520 @@
...
+      <Form.Label __className__="FormLabel">Email</Form.Label>
+      <Form.Control type="email" />
+    </Form.Field>
+  </Form.Root>
+);
+```
+
+```css line=2
+/* styles.css */
+.FormLabel[__data-invalid__] {
```
andy-hook, line 400: For completeness can we add `data-valid` with `color: green`?

data/primitives/components/form/0.0.1.mdx
```diff
@@ -0,0 +1,520 @@
...
+  color: red;
+}
+```
+
+### Accessing the validity state for more control
+
+Sometimes, you may need to access the raw validity state of a field in order to display your own icons, or interface with your own component library by passing certain props to it. You can do this by using the `Form.ValidityState` part:
```
andy-hook, line 407:
```suggestion
You may need to access the raw validity state of a field in order to display your own icons, or interface with a component library via it's defined props. You can do this by using the `Form.ValidityState` part:
```

data/primitives/components/form/0.0.1.mdx
```diff
@@ -0,0 +1,520 @@
...
+
+```jsx line=3-4,7-8
+<Form.Field name="name">
+  <Form.Label>Full name</Form.Label>
+  <Form.ValidityState>
+    {(validity) => (
+      <Form.Control asChild>
+        <TextField variant="primary" state={getTextFieldState(__validity__)} />
```
andy-hook, line 415:
```suggestion
        <TextField variant="primary" error={__validity__.valueMissing} />
```

I feel like this line is too opaque and we should be clearer on what's happening in the implementation. My suggestion is not the best but do you think it would help to be more explicit on what `getTextFieldState` is doing?

data/primitives/components/form/0.0.1.mdx
```diff
@@ -0,0 +1,520 @@
...
+      </Form.Control>
+    )}
+  </Form.ValidityState>
+</Form.Field>
+```
+
+### Server-side validation
+
+The component also supports server-side validation using the same `Form.Message` component.
+You can re-use the same messages you defined for client-side errors by passing a `forceMatch` prop which will for the message to show regardless of the client-side matching logic.
```
andy-hook, line 425:
```suggestion
You can re-use the same messages you defined for client-side errors by passing a `forceMatch` prop which will force the message to show regardless of the client-side matching logic.
```

data/primitives/components/form/0.0.1.mdx
```diff
@@ -0,0 +1,520 @@
...
+  const [serverErrors, setServerErrors] = React.useState({
+    email: false,
+    password: false,
+  });
+
+  return (
+    <Form.Root
+      onSubmit={(event) => {
+        // `onSubmit` only triggered if it passes client-side validation
+        event.preventDefault();
```
andy-hook, line 446: I find these two lines confusing, is the comment related to the `preventDefault`? it reads like it is yet the comment doesn't match my understanding of the `preventDefault` usage here.

data/primitives/components/form/0.0.1.mdx
```diff
@@ -0,0 +1,520 @@
...
+        const data = Object.fromEntries(new FormData(event.currentTarget));
+
+        // maybe do an async server call and gather some errors from the server to display
```
andy-hook, line 449:
```suggestion
        // Submit form data and catch errors in the response
```

data/primitives/components/form/0.0.1.mdx
```diff
@@ -0,0 +1,520 @@
...
+
+      <Form.Submit>Submit</Form.Submit>
+    </Form.Root>
+  );
+}
+```
+
+You should clear the server errors using the `onClearServerErrors` callback prop on the `Form.Root` part.
+
+> This will clear the server errors before the form is re-submitted, and when the form is reset.
```
andy-hook, line 495: Would this read better as part of the original paragraph?

data/primitives/components/form/0.0.1.mdx
```diff
@@ -0,0 +1,520 @@
...
+
+On top of this, it means you have control over when to reset single server errors.
```
andy-hook, line 497:
```suggestion
In addition, this provides control over when to reset single server errors.
```

data/primitives/components/form/0.0.1.mdx
```diff
@@ -0,0 +1,520 @@
...
+  </Form.Message>
+</Form.Field>
+```
+
+## Accessibility
+
+The component follows the "inline errors" pattern for validation:
+
+- Label and control are associated using the `name` provided on `Form.Field`
+- When one or more client-side error message display, they are automatically associated to their matching control and announced accordingly
```
andy-hook, line 519:
```suggestion
- When one or more client-side error messages display, they are automatically associated with their matching control and announced accordingly
```

andy-hook (commented):
This is looking great! I've added mostly nits and small changes. Let me know what you think or if any suggestions are not appropriate.

data/primitives/components/form/0.0.1.mdx
benoitgrelard: old copy/paste

data/primitives/components/form/0.0.1.mdx
benoitgrelard: I'm not sure it would to be honest, and I'm not sure you need to know what `getTextFieldState` does, the important thing is that it's `state = fn(validity)` so `validity` is needed as a prop somehow.

data/primitives/components/form/0.0.1.mdx
benoitgrelard: Yeah I can probably remove the comment.

benoitgrelard: I've reworked it differently.

data/primitives/components/form/0.0.1.mdx
andy-hook: Yeah you're probably right, I'm might be over thinking what we need to be demonstrating.

PR #628, opened 2023-05-05
Add a new page dedicated to the asChild prop

### LINKED ISSUE #269: Dedicated `asChild` page
We have had a few reports that have been assumed to be an issue with the new `asChild` API but also will have been an issue with the `as` prop. It seems the explicitness of the `asChild` API has made some of these issues more apparent for people so it might be useful to provide some documentation on these pitfalls and the solutions to them.

The most recent report we had was here https://github.com/radix-ui/primitives/issues/953

It's not clear what props the child should accept (or that props need to be forwarded). For now we're saying it should always accept DOM props. Ideally we would improve the DX with TS but TS doesn't currently support the ability to type the props of the child React Element.

stdavis:
	- http://localhost:3000/docs/primitives/overview/as-child-prop

This pull request:

- [ ] Fixes a bug
- [ ] Other

This is just a stab at #269. I ran into the issue of not spreading props within my component today so I thought that I would try and help get better docs up for it. I'm happy to make any changes that are requested. Thanks for the great project!

Closes #269


benoitgrelard:
Hey @stdavis thanks for getting this going. That's a great start! I'll take it from here and ship it. 🙏

data/primitives/guides/composition.mdx
```diff
@@ -0,0 +1,110 @@
...
+
+```jsx line=5
+// before
+const MyButton = () => <button />;
+
+// after
+const MyButton = (props) => <button {...__props__} />;
+```
+
+We recommend always doing it so that you are not concerned with implementation details (ie. which props/events to accept). This is also generally good practice anyway for "leaf" components.
```
andy-hook, line 58:
```suggestion
We recommend always doing this so that you are not concerned with implementation details (ie. which props/events to accept). We find this is good practice for "leaf" components in general.
```

data/primitives/guides/composition.mdx
```diff
@@ -0,0 +1,110 @@
...
+const MyButton = __React.forwardRef__((props, __forwardedRef__) => (
+  <button {...props} __ref__={__forwardedRef__} />
+));
+```
+
+Whilst this isn't necessary for **all** parts, we recommend always doing it so that you are not concerned with implementation details. This is also generally good practice anyway for leaf components.
+
+## Composing multiple primitives
+
+`asChild` can be used as deeply as you need to. This means it is a great way to compose multiple primitive's behaviour together.
```
andy-hook, line 82:
```suggestion
`asChild` can be used as deeply as you need to. This means it is a great way to compose multiple primitive's behavior together.
```

andy-hook (commented):
This is excellent, I've added some comments but feel free to ignore any that you feel are not a decent improvement.

benoitgrelard (approved):

stdavis:
Thanks for pushing this through! These docs look great and super-useful to those who are new to Radix like me. 👍

PR #644, opened 2023-06-05
fix(scroll-area): Tailwind classname

bukinoshita:

This pull request:

- [ ] Fixes a bug
- [ ] Other

It updates the arbitrary value to use the default Tailwind value


joaom00:
@benoitgrelard I think is fine in that case as `max-w-11` maps to `44px` https://tailwindcss.com/docs/customizing-spacing#default-spacing-scale

benoitgrelard (commented):
Hey Bu! I believe we wanted to keep all the TW demos looking exactly the same, hence the use of arbitrary values in all TW demos.

PR #655, opened 2023-07-14
[RadioGroup] Removed data-unchecked in RadioGroupIndicator

onmax:
Since this component does only rendered when the item is checked, it would be just 'checked' or component not found in the DOM.

This pull request:

- [ ] Fixes a bug
- [ ] Other

### DIFF
--- a/data/primitives/components/radio-group/1.1.3.mdx
+++ b/data/primitives/components/radio-group/1.1.3.mdx
@@ -269,7 +269,7 @@ Renders when the radio item is in a checked state. You can style this element di
   data={[
     {
       attribute: '[data-state]',
-      values: ['checked', 'unchecked'],
+      values: ['checked'],
     },
     {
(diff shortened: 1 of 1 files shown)


benoitgrelard:
What is there is actually correct, it's not `data-checked` or `data-unchecked` but rather `data-state="checked"` or `data-state="unchecked`.

onmax:
Maybe I am wrong, but here I am talking about the "span" element inside the button.

image

When it is checked the RadioIndicator will be inserted inside the button and it will have `data-state="checked"`.

When it is unchecked, the RadioIndicator will be destroyed.

Therefore, RadioIndicator only has one state.

It is a stupid little detail though.

benoitgrelard:
I see, it still has 2 states because so it can be animated with CSS animations (unmount animation) or also when using `forceMount` to animate with JavaScript.

PR #676, opened 2023-08-11
Fix fonts in portalled components examples

vladmoroz:
- Make sure examples use default fonts, including in the portalled components
- Make sure other elements use Radix brand fonts
- Do a pass on monospace fonts (weren’t configured consistently between Radix Themes and existing Modulz design system usage)


pages/styles.css
```diff
@@ -0,0 +1,57 @@
...
+}
+
+#__next {
+  position: relative;
+  z-index: 0;
+}
+
+#__next > .radix-themes {
+  min-height: 100vh;
+}
```
andy-hook, line 42: Is there a reason this was moved? it's a more brittle selection.

pages/styles.css
```diff
@@ -0,0 +1,57 @@
...
+
+.radix-themes.radix-themes-custom,
+.radix-themes-custom .radix-themes:not(.radix-themes-example) {
```
andy-hook, line 45: How do you feel about this naming?

`radix-themes-custom-font`
`radix-themes-default-font`

andy-hook, line 45: Should we colocate these inside of `themes-config.css`?

vladmoroz: Let's do this
`radix-themes-custom-fonts`
`radix-themes-default-fonts`

vladmoroz: There's no `themes-config.css` anymore (https://github.com/radix-ui/website/pull/676#discussion_r1291433670)

pages/styles.css
vladmoroz: Updated in c79ab8d64cd4a208e8b34b7b21e0b14d97fb34f0

pages/styles.css
vladmoroz: Updated in c79ab8d64cd4a208e8b34b7b21e0b14d97fb34f0

andy-hook: Right! missed the deletion

andy-hook (approved):

PR #684, opened 2023-08-15
Target Node 18

andy-hook:
We were previously targeting `14.x` which Vercel deprecated support for as of today (Aug 15, 2023) causing all builds to fail.

CleanShot 2023-08-15 at 13 16 53

https://vercel.com/docs/concepts/functions/serverless-functions/runtimes/node-js#node.js-version

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,10 +2,13 @@
   "name": "radix-site",
   "version": "0.0.0",
   "private": true,
+  "engines": {
+    "node": "18.x"
+  },
   "scripts": {
-    "dev": "next dev",
-    "build": "next build",
-    "start": "next start",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -2,10 +2,13 @@
...
   "version": "0.0.0",
   "private": true,
+  "engines": {
+    "node": "18.x"
+  },
   "scripts": {
-    "dev": "next dev",
-    "build": "next build",
-    "start": "next start",
+    "dev": "export NODE_OPTIONS=--openssl-legacy-provider && next dev",
```
andy-hook, line 9: Temp workaround to avoid upgrading many dependencies.

See this change for those deps: https://github.com/radix-ui/website/pull/685

vladmoroz (approved):

PR #686, opened 2023-08-15
[fix] Change the regex that extracts the alpha value from 'hsla', as the previous one was not doing it correctly

Azuky:

This pull request:

- [x] Fixes a bug
- [ ] Other

Preview URL: https://radix-website-git-fork-azuky-fix-hsla-to-alpha-workos.vercel.app/colors/docs/palette-composition/scales

If you copy the object as hex or as SVG, it now correctly copies the alpha values. Previously, these values were ignored

### DIFF
--- a/components/Scale.tsx
+++ b/components/Scale.tsx
@@ -6,6 +6,8 @@ import tinycolor from 'tinycolor2';
 
 type Scale = Record<string, string>;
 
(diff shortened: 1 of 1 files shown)


andy-hook (approved):
Works great! thank you.

PR #688, opened 2023-08-16
Remove @modulz/design-system

vladmoroz:
- Finish removing @modulz/design-system
- Use Untitled Sans files which don’t have the macOS @1x hinting problem
- Declare Untitled Sans Medium as "bold" ("medium" is kept for backwards compatibility in primitive demos until we do a pass on them at some point)
- Make sure styles that still use Stitches can be SSR'd
- Visual tweaks on Primitives homepage


package.json
```diff
@@ -18,7 +18,7 @@
   "dependencies": {
     "@algolia/autocomplete-core": "^1.7.1",
     "@algolia/autocomplete-preset-algolia": "^1.7.1",
-    "@modulz/design-system": "1.0.0",
+    "@mdx-js/loader": "1.6.22",
```
andy-hook, line 21: Was this previously removed in error? how is this package used?

vladmoroz: Not sure how it ended up here, it had sneaked in during git conflict resolution. Removed

andy-hook:
Do we need to tweak any other ddetails given the change to glyph scaling? here are the primitives feature lists on my x1, they feel pretty tight now: 

CleanShot 2023-08-16 at 17 33 51

vladmoroz:
@andy-hook I prefer it with the font being a hair tighter tbh, but it could be somewhat different on your screen. We also didn't tweak anything when updating it in WorkOS UIs. Could you do before/after screenshots of what you are seeing?

Here's what I'm seeing, before:

Screenshot 2023-08-16 at 19 04 27


***

After:

Screenshot 2023-08-16 at 19 04 23

andy-hook:
before/after

List:
Frame 3

Longform:
Frame 2 (3)

andy-hook:
> We also didn't tweak anything when updating it in WorkOS UIs

You're right, and I remember now having a similar discussion then. The nature of the issue means that we can't tweak on one density without affecting the other anyway.

The new long form looks good to me, it's only the feature lists that feels like they need a touch more margin between the items.

Regardless, I think we should move forward as this is the same discussion as last time 😄

andy-hook (approved):

vladmoroz:
> it's only the feature lists that feels like they need a touch more margin between the items

Yeah these look overly tight @2x too, I think the issue was just a bit masked @1x before. Making the gap a bit larger:

image

PR #689, opened 2023-08-17
Fix innacurate Hover Card docs

TGlide:
This pull request:

- [ ] Fixes a bug
- [ ] Other

The current docs state that Hover Card only works on hover. This is no longer the case, as focus also triggers it.


benoitgrelard (approved):
Thanks for the headstart here, I've made a few more subtle changes.

PR #698, opened 2023-09-02
fix: Delete duplicated onValueChange prop in Accordion component

HassenIO:
## Problem

In the documentation of the Accordion component, the `onValueChange` root prop is cited twice.

https://www.radix-ui.com/primitives/docs/components/accordion#root

## Solution

In this PR, I remove the unnecessary second mention.

## Future Improvement

In another PR, order the props in alphabetical order. This will:

- Avoid other/future duplication
- Ease the search of a property for a component

### This pull request is about:

- [ ] Fixes a bug
- [ ] Other

### DIFF
--- a/data/primitives/docs/components/accordion/1.1.2.mdx
+++ b/data/primitives/docs/components/accordion/1.1.2.mdx
@@ -154,18 +154,6 @@ Contains all the parts of an accordion.
       ),
     },
     {
-      name: 'onValueChange',
-      required: false,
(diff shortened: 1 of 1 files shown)


benoitgrelard:
Hi @HassenIO this isn't a duplicate, if you look closely the signature is different. This is because the accordion works differently base on what `type` is passed in (single/multiple).

I agree this is not ideal and we should think of a different way to represent props like these a bit differently. At the moment this is our best solution.

PR #705, opened 2023-09-13
docs(navigation-menu): tsx instead of jsx

konhi:
typescript is an industry standard, so it makes sense to use types in code example, so the user don't have to add them manually.

---

This pull request:

- [ ] Fixes a bug
- [ ] Other


andy-hook:
Thanks for this, though if we ever decide to provide TS examples I think it'd be best to provide them consistently for all components.

Going to close for now as I can't see us having the bandwidth to do so in the near feature.

PR #737, opened 2024-01-29
Update .nvmrc to match engines.node in package.json

jameshfisher:
Fixes bug:

1. `git clone git@github.com:radix-ui/website.git`
2. `cd website`. Node version is changed to v14, following `.nvmrc`
3. `yarn` gives output:

```
yarn install v1.22.18
[1/5] 🔍  Validating package.json...
error radix-site@0.0.0: The engine "node" is incompatible with this module. Expected version "18.x". Got "14.15.0"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```


.nvmrc
```diff
@@ -1 +1 @@
-14.15.0
+18.15.0
```
JeromeFitz, line 1: Nice, thanks for the update @jameshfisher ! 🎉 

Current Node 18 LTS is `18.19.0` (ref: https://nodejs.org/en/blog/release/v18.19.0)

Thoughts on either updating to that specific version or just `18`?

If it is `18` it will use whatever `node@18` version is already on the local system (or install latest if does not exist).

benoitgrelard, line 1:
```suggestion
18
```

benoitgrelard (approved):

PR #759, opened 2024-04-02
Radix Themes: Improve Select docs

vladmoroz:
Add the commonly requested documentation on preventing an SSR layout shift and rendering an icon in the trigger

### DIFF
--- a/data/themes/docs/components/select.mdx
+++ b/data/themes/docs/components/select.mdx
@@ -287,3 +287,56 @@ Set `position="popper"` prop to position the select menu below the trigger.
   </Select.Content>
 </Select.Root>
 ```
+
+### With SSR
+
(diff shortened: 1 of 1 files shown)


data/themes/docs/components/select.mdx
```diff
@@ -287,3 +287,56 @@ Set `position="popper"` prop to position the select menu below the trigger.
   </Select.Content>
 </Select.Root>
 ```
+
+### With SSR
+
+When using server-side rendering, you might notice a layout shift after hydration. This is because Trigger executes client-side code to display the label that corresponds to the resolved value. To avoid that layout shift, you can render the Trigger label manually by mapping values.
```
benoitgrelard, line 293: Avoiding to use "label" as it's easily confused for an actual label:

```suggestion
When using server-side rendering, you might notice a layout shift after hydration. This is because Trigger executes client-side code to display the selected item's text content. To avoid that layout shift, you can render it manually by mapping values.
```

vladmoroz: Good call, updated

benoitgrelard (approved):

PR #773, opened 2024-05-02
Adjust the ExampleThemes* page layout

danilo-leal:

This pull request:

- [x] Fixes a bug
- [ ] Other

---

Hey y'all 👋 This is not a functional bug per se, but more of a design issue that I noticed while browsing through these ExampleThemes* components/pages on a small viewport: they bleed off of it, get a horizontal scroll, and the background/title get off-center. The video below illustrates all of this better + the fix (which was just adding `wrap` to the Flex parent containers):

https://github.com/radix-ui/website/assets/67129314/d7f18215-7906-4109-87a2-f7da494286db

### DIFF
--- a/components/ExampleThemesDashboard.tsx
+++ b/components/ExampleThemesDashboard.tsx
@@ -61,7 +61,7 @@ export const ExampleThemesDashboard = ({ focusable = true, ...props }: ExampleLa
   });
 
   return (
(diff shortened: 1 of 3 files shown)


chaance:
I think this is by design, though ideally our examples should be responsive. I think we should consider revisiting this. I'll add it to our backlog.

PR #775, opened 2024-05-06
Update Tailwind demos

ryanleichty:
There were a number of inconsistencies that I noticed when using the Tailwind examples, so this PR makes everything more consistent.

- Updated a number of redundant arbitrary classes with native Tailwind classes.
- Sorted all tailwind classes using the `prettier-plugin-tailwindcss`.
- Fixed a few incorrect classes in the process of sorting (e.g.: `color-white` to `text-white`). The `prettier-plugin-tailwindcss` plugin moves unrecognized classes to the beginning of the string which can help find incorrect classes.

It might be worth considering adding the Radix design system within the Tailwind config files versus using arbitrary values for the spacing scale. I found a number of old design system values, but wasn't 100% positive which were used for the primitives when they were first created.

This pull request:

- [ ] Fixes a bug
- [ ] Other


chaance:
It would definitely be nice to fix any inconsistencies or incorrect use, but I'd prefer to separate the more opinionated use of `prettier-plugin-tailwindcss`. It makes the diffs a bit hard to tell what is being re-ordered vs. what actual issues we're fixing.

PR #776, opened 2024-05-07
Upgrade problematic dependencies

vladmoroz:
- Upgrade Next.js
- Upgrade most of the MDX deps
- Add missing dependencies
- Remove inlined colorjs.io

Fixes an issue when Next.js/Webpack minification would bust colorjs.io when built


utils/color.js
benoitgrelard, line 1: was that a copy of the lib?

benoitgrelard (approved):
Yep, it works well now, nice one!

utils/color.js
vladmoroz: Yes, it was a temporary workaround to a similar issue when the color generator was broken in prod, but not locally. (Back then I didn't have the time to dig into the root cause when we were shipping the color generator)

PR #847, opened 2025-02-06
docs(direction-provider): update Direction Provider example to use a named import

AntanasMisiunas:
This pull request:

- [ ] Fixes a bug
- [ ] Other

Given the [installation instructions](https://www.radix-ui.com/primitives/docs/utilities/direction-provider#installation):
```sh
npm install @radix-ui/react-direction
```

It doesn't make sense to showcase an example importing from the entire Radix UI package.

```tsx
import { Direction } from "radix-ui";
```


chaance:
This by design. We are now recommending users install and import from the `radix-ui` package, as it is and simpler to maintain version consistency between Radix packages.

